PR #9, opened 2026-07-04
docs(readme): fix Library use example — package name and renderTextToImages

### LINKED ISSUE #8: README `Library use` example uses a non-existent import + function
The `Library use` snippet in the README doesn't run as written:

```ts
import { renderTextToPngs, transformAnthropicMessages } from "pxpipe";
```

Two things are off:

1. The published package is `pxpipe-proxy` (the `name` in `package.json`), not `pxpipe`.
2. The root export (`src/core/index.ts`) is `renderTextToImages` — `renderTextToPngs` isn't re-exported there, so the named import fails.

`renderTextToImages` also returns `{ pages, droppedChars, pixels }`, so the follow-up line (`const imgs = await renderTextToPngs(...)  // RenderedImage[]`) doesn't match the real shape either.

Small, but it's the first code a library user copies. Happy to send a PR.

rldyourmnd:
Fixes #8.

The `Library use` example imported `renderTextToPngs` from `"pxpipe"`, but the package is `pxpipe-proxy` and the root barrel (`src/core/index.ts`) exports `renderTextToImages`, not `renderTextToPngs` — so the snippet failed on both the specifier and the function name.

- import from `pxpipe-proxy`
- use `renderTextToImages`, and show its real return shape (`const { pages } = ...`, `pages[i].png`)

Docs-only; `pnpm run typecheck && pnpm test && pnpm run build` stay green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,9 +117,9 @@ sparse prose stays text. Events log to `~/.pxpipe/events.jsonl`.
 ## Library use (no proxy)
 
 ```ts
-import { renderTextToPngs, transformAnthropicMessages } from "pxpipe";
+import { renderTextToImages, transformAnthropicMessages } from "pxpipe-proxy";
 
-const imgs = await renderTextToPngs(toolResultText);            // RenderedImage[]
+const { pages } = await renderTextToImages(toolResultText);     // pages[i].png: Uint8Array
 const { body, applied, info } = await transformAnthropicMessages({
   body: requestBytes,
   model: "claude-fable-5",


teamchong (approved):
Thanks @rldyourmnd

rldyourmnd:
you too for this idea!)

thurtea:
Screenshot 2026-07-04 152714
Screenshot 2026-07-04 152723
I made a website where I can take the prompts and it makes the image using your work. Thank you I will test this

PR #13, opened 2026-07-04
fix(export): apply include/exclude and size limit to --git untracked files

### LINKED ISSUE #12: `pxpipe export --git` ignores --include/--exclude and the size cap for untracked files
In `--git` mode, untracked files are collected in `collectSource` (`src/node.ts`) with only a binary check:

```ts
const untrackedOut = gitRun(['ls-files', '--others', '--exclude-standard'], cwd) ?? '';
for (const rel of untrackedFiles) {
  const full = path.join(cwd, rel);
  let buf; try { buf = fs.readFileSync(full); } catch { continue; }
  if (!looksLikeBinary(buf)) { untracked += `\n===== ${rel} =====\n` + buf.toString('utf8'); }
}
```

Directory mode (`walkDir`) and single-file mode both also apply `shouldIncludeFile(rel, include, exclude)` and the `MAX_FILE_BYTES` (1 MiB) cap. So on the untracked path:

- `pxpipe export --git --include "*.ts"` still pulls in every untracked file regardless of the filter
- an untracked multi-MB file is read fully into memory (no size cap)

The missing size cap is the one I'd call a real hazard. PR incoming.

rldyourmnd:
Fixes #12.

`--git` untracked collection only applied the binary check, unlike directory / single-file modes which also apply `--include`/`--exclude` and the 1 MiB `MAX_FILE_BYTES` cap. So `export --git --include "*.ts"` ignored the filter, and an oversized untracked file was read whole.

- extracted the shared gate into `src/export-collect.ts` (`readExportTextFile()` — include/exclude → size → binary). It's a small import-safe module: `src/core/export.ts` is contractually fs-free and `src/node.ts` starts the server on import, so neither was a good home for a unit-testable fs helper
- routed all three collection paths through it; untracked skips now warn like explicit targets do
- `tests/export-collect.test.ts` covers include, exclude, oversized, the size boundary, binary, inaccessible

Verified end-to-end: `export --git --include "*.ts"` in a throwaway repo now excludes an untracked `.md`, skips an oversized `.ts` (with a warning), and keeps the `.ts`. typecheck/test/build green.

### DIFF
--- a/src/export-collect.ts
+++ b/src/export-collect.ts
@@ -1,0 +1,67 @@
+/**
+ * Shared file-reading gate for `pxpipe export`. Every collection mode
+ * (directory walk, explicit single-file target, and `--git` untracked files)
+ * must apply the SAME three checks — include/exclude globs, a max size, and a
+ * binary sniff — so the untracked path can no longer read gigabyte or filtered
+ * files that directory mode would have skipped.
+ *
+ * Kept in its own import-safe module (no top-level side effects, unlike
+ * src/node.ts which starts the server on import) so it can be unit-tested
+ * directly. src/core/export.ts is deliberately fs-free, so the fs-touching gate
+ * lives here rather than there.
+ */
+
+import * as fs from 'node:fs';
+import { shouldIncludeFile } from './core/export.js';
+
+/** Files larger than this are skipped by export (1 MiB). A single export bundle
+ *  is meant to be paste-sized; a multi-MB file is never intentional context and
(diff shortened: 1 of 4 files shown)


rldyourmnd:
Pushed a follow-up: an end-to-end `pxpipe export --git` test in a throwaway git repo (tracked baseline + untracked keep.ts / large skip.md / oversized huge.ts / binary bin.ts). The original bug lived in the `--git` collectSource path, not the extracted helper the unit tests cover, so this exercises the real CLI end to end.

PR #17, opened 2026-07-04
fix(openai): record outgoingTextChars for Responses transforms

### LINKED ISSUE #16: Responses transform never sets info.outgoingTextChars (regression denominator missing)
`transformOpenAIChatCompletions` sets `info.outgoingTextChars = countOutgoingTextChars(req)` right before returning, but `transformOpenAIResponses` sets `info.compressed = true` and returns without it.

`outgoingTextChars` is documented as the denominator for the `tokens ≈ α·outgoingTextChars + β·imagePixels` regression, so compressed Responses rows carry no denominator — Chat and Responses telemetry are inconsistent. PR incoming.

rldyourmnd:
Fixes #16.

The Chat path set `info.outgoingTextChars = countOutgoingTextChars(req)`; the Responses path didn't set it at all, so compressed Responses rows silently carried no regression denominator.

- add `countResponsesOutgoingTextChars()`: instructions + message-item text (string or `input_text` parts, via `responsesContentText` so `input_image` base64 is excluded) + flat tool name/description/parameters
- set it on the Responses success path, mirroring Chat
- test asserts `outgoingTextChars > 0` for a compressed Responses request

typecheck/test/build green.

### DIFF
--- a/src/core/openai.ts
+++ b/src/core/openai.ts
@@ -418,6 +418,32 @@ function countOutgoingTextChars(req: OpenAIChatRequest): number {
   return n;
 }
 
+/** Outgoing text-char denominator for the GPT Responses regression, mirroring
+ *  countOutgoingTextChars for Chat: instructions + message-item text (string or
+ *  input_text parts) + flat tool name/description/parameters. input_image base64
+ *  is excluded on purpose — it is image cost, not text (responsesContentText
+ *  already drops non-text parts). */
+function countResponsesOutgoingTextChars(req: ResponsesRequest): number {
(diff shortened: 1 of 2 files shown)


rldyourmnd:
Pushed a follow-up strengthening the assertion from `outgoingTextChars > 0` to a structural check: with a real rendered image, pointer, prompt, and rewritten tool in the output, it verifies the count includes the text (instructions/parts/tool fields) and stays well below the input_image base64 length — i.e. the base64 is excluded from the denominator.

PR #19, opened 2026-07-04
fix(openai): stop double-billing native GPT tool descriptions

### LINKED ISSUE #18: GPT tool descriptions are double-billed: kept native AND rendered into the image
On the GPT path, `renderToolDoc` / `renderFlatToolDoc` render the tool `description` into the imaged slab:

```ts
const parts = [`## Tool: ${f.name ?? '?'}`];
if (typeof f.description === 'string' && f.description.length > 0) parts.push(f.description);
```

But `rewriteToolsForGpt` / `rewriteFlatToolsForGpt` keep `function.description` native — they only strip the schema annotations. So the description goes out **twice**: as native JSON text and as image pixels.

Meanwhile `gptBaselineImagedTokens` credits only `orig − stripped` (the schema-annotation delta); the native-kept description cancels, so it's never counted as "saved". Net effect: the image pays for bytes nothing was saved on, and a big-description / tiny-schema tool reports `compressed=true` while the description is still billed natively. (The Anthropic path is fine — it stubs the native description, so imaging it there is a real move, not a duplicate.)

Two ways to make the accounting honest:

- **A** — stub the native description, keep the full text only in the image (real compression, but changes what goes to OpenAI)
- **B** — stop imaging the description (it stays native); image only the schema, which carries the annotations actually stripped

I went with **B** in the PR (conservative — native JSON untouched, and the README says GPT tools stay native), but happy to switch to A if you'd prefer.

rldyourmnd:
Fixes #18 (Design B — conservative: native `tools[]` untouched).

`renderToolDoc`/`renderFlatToolDoc` imaged the tool description, but `rewriteToolsForGpt` keeps `function.description` native (only schema annotations are stripped). So each description was billed twice — native text + image pixels — while `gptBaselineImagedTokens` credits only the stripped-schema delta (in which the native-kept description cancels).

- `renderToolDoc`/`renderFlatToolDoc` no longer push the description; they image the heading + schema (which carries the stripped annotations), so every imaged byte maps to something actually removed from native
- `gptBaselineImagedTokens` math is unchanged (it was already correct); fixed its comment, which called the delta "description tokens" when it's schema-annotation tokens
- tests: the "no instruction context" cases now use a realistic big **schema** (the profitable case under the dedupe); the task-tool strip regressions get a system/instructions slab so compression still triggers; added a test that a big description over a tiny schema is **not** imaged (no double-bill) and passes native through untouched

If you'd rather move the description into the image and stub it native (Design A — more compression, but changes the outgoing tool text), I'm happy to redo it that way. typecheck/test/build green.

### DIFF
--- a/src/core/openai.ts
+++ b/src/core/openai.ts
@@ -322,19 +322,26 @@ function isFlatFunctionTool(tool: unknown): tool is ResponsesFlatTool {
  *  IMAGED, so carrying the schema here is compression, not duplication: the imaged
  *  copy keeps param docs readable while tools[] ships the stripped skeleton.
  *  (Contrast transform.ts renderToolDoc: text reference → prose only.) */
+// On the GPT path the native tools[] KEEP their description — only the schema
+// annotations are stripped (see rewriteToolsForGpt). So the description must NOT
+// be imaged: doing so double-bills it (native text + image pixels) while the
+// savings baseline credits only the stripped-schema delta. Image the heading +
+// full schema (which carries the stripped annotations); the description rides
+// natively. (The Anthropic path differs on purpose — it stubs the native
+// description and images the full doc, a real move rather than a duplicate.)
 function renderToolDoc(tool: OpenAIFunctionTool): string {
   const f = tool.function;
   const parts = [`## Tool: ${f.name ?? '?'}`];
-  if (typeof f.description === 'string' && f.description.length > 0) parts.push(f.description);
   if (f.parameters !== undefined) {
     parts.push('```json\n' + JSON.stringify(f.parameters) + '\n```');
   }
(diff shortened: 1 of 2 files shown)


rldyourmnd:
Pushed a follow-up. The `CHAT_HEADER` / `RESPONSES_HEADER` still told the model the images hold "full tool/schema documentation" — but after this dedupe they hold the SCHEMA only; the description stays native. Corrected the header + pointer wording and added a regression test that the framing never claims "full tool" and keeps native definitions authoritative.

teamchong:
Thanks for this @rldyourmnd, closing as I believe the core change already landed in `main` via 4c78363

PR #27, opened 2026-07-04
fix(vision): patch-based Anthropic image-token model + doc sync

### LINKED ISSUE #26: Anthropic vision cost uses the retired w×h/750 formula, not the documented 28-px patch count
The Anthropic image-token cost in `src/core/transform.ts` / `src/core/export.ts` uses `ceil(w*h/750*1.10)` (`ANTHROPIC_PIXELS_PER_TOKEN=750`). Anthropic's current docs no longer document that formula — they document a **28-px patch** model: an image costs `⌈width/28⌉ × ⌈height/28⌉` visual tokens, computed after downscaling to the model tier's limits, with a reference `count_image_tokens` implementation. (`/750` was a ~4–5% continuous approximation of the same 28²=784 px²/patch grid, and it ignored the tier downscale, so it grossly overcharged large images.)

Tiers (from the docs' resolution table):
- **high-res** (Fable 5, Mythos 5, Opus 4.8, Opus 4.7, Sonnet 5): max long edge 2576 px, max 4784 visual tokens
- **standard** (all others): 1568 px / 1568 tokens

Refs: https://platform.claude.com/docs/en/build-with-claude/vision and .../vision-coordinates

Two smaller drifts tie in:
- `transform.ts` `DEFAULTS.cols = 313` (→ 1573 px) overshoots `render.ts` `MAX_WIDTH_PX = 1568` (a 0.997× resample); `DEFAULT_COLS` is 312.
- README / RENDER_SIZING / TRANSFORM_INFO still describe the old 1928×1928 / 384-col / 1932-height / ~92k-char geometry and the `/750` formula, and RENDER_SIZING calls `shrinkColsToContent` a no-op (it delegates to `measureContentCols` now). Thresholds are documented as 1000/2000 but the code uses 6000/6000.

PR incoming that grounds the cost model on the patch formula and syncs the docs.

rldyourmnd:
Fixes #26.

Grounds the Anthropic image-token cost on the current documented **28-px patch** model and syncs the docs. Verified against Anthropic's primary docs (patch formula + tier table + reference `count_image_tokens`, reproduced exactly).

- `src/core/anthropic-vision.ts` (new): `patchTokens`, `anthropicVisionProfile(model)`, `anthropicVisionTokens(model,w,h)` with the documented long-edge + token-budget downscale (standard 1568/1568, high-res 2576/4784). Matches Anthropic's worked cost table (e.g. 1928² high-res → 4761; 1928² standard → 1521; 1568×728 → 1456).
- gate (`imageTokensForRows`) now counts per-image 28-px patches; the 10% conservatism is renamed `ANTHROPIC_GATE_MARGIN` and kept ONLY at the gate — the documented formula carries no margin. **pxpipe pages are ≤ 1568×728, so no tier downscale ever applies to proxy output and the net gate shift is ~4%.**
- export (`exportImageTokens`) routes claude models through `anthropicVisionTokens` (accurate, un-margined reporting).
- `DEFAULTS.cols` 313 → 312 (1568 px == `MAX_WIDTH_PX`); fix `shrinkColsToContent`'s stale "no-op" docstring and `render.ts`'s stale page-cap header.
- docs (README, RENDER_SIZING, TRANSFORM_INFO, legibility audit) synced to 1568×728 / 312-col / 28k-char geometry, 6000/6000 thresholds, and the patch billing model.
- `tests/anthropic-vision.test.ts` covers the patch math, tiers, and downscale; export tests updated off the `/750` numbers.

Full suite green (638 on this branch); typecheck + build green. (I also verified the whole PR set merges and stays green together — see the integration note.)

### DIFF
--- a/docs/LEGIBILITY-AUDIT-2026-07-01.md
+++ b/docs/LEGIBILITY-AUDIT-2026-07-01.md
@@ -13,6 +13,11 @@ follow from it.
   ~2.8×4.4 px. We were paying full price for pixels destroyed in transit.
 - **Shipped:** page geometry clamped to **1568×728 = 1.14 MP** (WYSIWYG: billed/actual
   pixel ratio 3.25 → 1.04). No token-cost change per image; 614/614 tests green.
+- **Update (2026-07-04):** the `px/750` figure above is a ~4–5% continuous
+  approximation. Anthropic's current docs bill the exact **28-px patch count**
+  `⌈w/28⌉×⌈h/28⌉` (a 1568×728 page = 56×26 = **1456** tokens, ≈ the px/750 slope
+  measured here). The gate and export now use that exact formula via
+  `src/core/anthropic-vision.ts` (tiers: standard 1568/1568, high-res 2576/4784).
 - Reading exact strings off even **crisp** 5×8 tops out at **63%** (24-token blind
   test). Every miss is one of two classes the glyph-confusability matrix predicts:
   **case-normalization** (camelCase) and **single-glyph substitution** (hex/num).
--- a/docs/RENDER_SIZING.md
+++ b/docs/RENDER_SIZING.md
@@ -64,9 +64,9 @@ The profitability gate resolves the same profile as the renderer and derives:
 - per-image token cost
 
 Changing cell spacing or font therefore cannot leave cost prediction on a
-hardcoded 5×8 canvas. Claude uses the Anthropic pixel formula with a conservative
-margin, GPT uses model-specific tile/patch pricing, and Grok uses the measured
-tokens-per-megapixel rate.
+hardcoded 5×8 canvas. Claude uses the Anthropic 28-px patch formula (with a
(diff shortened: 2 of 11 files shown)


rldyourmnd:
Pushed a follow-up making `anthropicVisionTokens` an exact translation of Anthropic's documented resize, not an approximation. The old continuous sqrt+decrement scale matched the common cost table but drifted on extreme aspect ratios (it scaled both edges by one factor and floored, instead of pinning the long edge and rounding the short one) — e.g. standard-tier 2510×7800 billed 1064, the code gave 1008. Now it uses `fitsTier()` + a binary search over the long edge (short edge = `round(long × aspect)`, swap-recurse for portrait), and the tests cover the full official cost table on both tiers, the three extreme-aspect counterexamples, and a 400-case property check vs an independent reference. Also fixed two stale comments (`~50k chars` → `~28k`; `px/750` → the exact 28-px patch count). Only the general-purpose estimator changes; pxpipe pages are ≤1568×728 so the resize never fires for proxy output.

PR #39, opened 2026-07-04
Opt-in context risk router: keep secrets & high-risk blocks as text

rYo-STUDIO-1bit:
## What & why

pxpipe's image lane is lossy by design, and the README is honest that exact strings can be silently mis-OCR'd. `factSheetText()` already rescues sparse anchors (paths/SHAs/versions/…) next to imaged blocks — that part is solved.

The gap this addresses is **secrets**. The factsheet extracts a secret's *key name* but not its value, and the renderer draws the full secret **value** into the PNG, where an allowlisted model (Fable-5 reads dense glyphs near-perfectly) can read it straight back. Nothing populates `keepSharp` by default, so a secret in a tool_result or system prompt is **silently imaged**. That's the one failure mode a compression proxy shouldn't have.

This is a small, **opt-in, off-by-default** router that keeps secrets (and, optionally, other high-risk/dense blocks) as text. It reuses your existing seams — **no transform pipeline fork**.

## Design

Deterministic, no model calls (per the "keep it testable" spirit of the existing gates):

- `src/core/exact-token-extractor.ts` — regex + Shannon-entropy extraction of paths/hashes/uuids/commands/urls/versions/secrets. **Secrets are masked at source** — a raw secret never enters the token list, logs, or a test snapshot.
- `src/core/risk-classifier.ts` — `assessContextRisk()`: size + anchor-density → `text_only | image_only | image_plus_exact_rescue | redact_or_block`. A density fallback keeps anchor-dense blocks as text (a rescue strip that reproduces most of the block would cost more than the text it replaces).
- `src/core/context-router.ts` — `makeKeepSharp(policy)`: a drop-in for your existing `TransformOptions.keepSharp` hook. Policies: `default | coding-agent | research | strict`.

## Wiring (off by default → byte-identical behavior when unset)

- `node.ts` reads `PXPIPE_CONTEXT_ROUTER`; when set, passes `keepSharp: makeKeepSharp(policy)` + `guardSlabSecrets: true` into the transform.
- `transform.ts` gains `guardSlabSecrets` (default `false`): when the static system slab contains a secret, keep the whole request as text (the slab path has no `keepSharp` check). **Fires only on secrets**, so a path/version-rich tool-doc slab still images — compression is untouched in the common case.

```
PXPIPE_CONTEXT_ROUTER=on        # coding-agent policy
PXPIPE_CONTEXT_ROUTER=strict    # any anchor stays text
PXPIPE_CONTEXT_ROUTER=off       # (default) no change
```

## Tests

22 passing (`tests/context-router.test.ts` 13 unit + `tests/context-router-e2e.test.ts` 9 e2e through the real `transformRequest`). The e2e tests prove, for both tool_results and the system slab: without the router the secret is imaged (not present as text); with it the block stays text and the secret survives verbatim; a safe block still images. `tsc --noEmit` clean. Full suite green except two pre-existing order-dependent e2e flakes unrelated to this change (`proxy-usage`, `design-behavior-e2e` — they fail the same way on `main` with these files stashed). Runnable demo: `npx tsx scripts/context-router-demo.ts`.

## Scope / honesty

Labeled a **prototype** in `docs/CONTEXT_ROUTER.md` — not claimed production-safe. Known limits documented there: secret detection is heuristic (prefix + entropy, conservative-false-positive by policy); the density gate is heuristic, not yet fed into `evalCompressionProfitability`; only the 3 live-region sites + the slab are guarded. Happy to adjust naming, policy defaults, or split this up however suits the repo.


rYo-STUDIO-1bit:
Added a savings replay (`scripts/context-router-bench.ts`, commit d422ced) to answer "does the router hurt savings?" — priced with pxpipe's own `evalCompressionProfitability`, ALL-TEXT vs IMAGE-EVERY vs ROUTER-ON:

| scenario | IMAGE-EVERY | ROUTER-ON | cost of safety | secret |
|----------|-------------|-----------|----------------|--------|
| **Typical turn** (no secret, sparse anchors) | 76.0% | **76.0%** | **0.0pp** | safe both ways |
| **Risky turn** (secret in a profitable log + a path-dense listing) | 75.4% — *images the secret* | 39.1% | 36.3pp | IMAGE-EVERY leaks it; ROUTER-ON safe |

So: **free in the common case** (no secret / no dense block → images exactly what pxpipe would). The cost only appears on risky turns, and it's dominated by the conservative choice to keep a *whole* block as text when it holds a secret (a 6.5k-token log with one secret line stays fully text).

The obvious next lever is **redaction**: mask just the secret *value* and image the rest — safe *and* compressed, recovering most of that 36pp. The lane is even named `redact_or_block` but currently only "blocks" (keeps text). Happy to build that if the overall direction looks right to you.

rYo-STUDIO-1bit:
Built the redaction lane (commit 0bd8173). Opt-in via `PXPIPE_CONTEXT_ROUTER=redact`: a secret span is masked to `[redacted-secret]` *before* rendering, so the block still images (savings preserved) with no secret in the pixels.

Wired through a new `TransformOptions.redactBlock` hook at all three live-region sites (reminder / tool_result / tool_result_part), consulted after `keepSharp` has seen the original. The redacted text is what gets gated, imaged, fact-sheeted, and recorded as recoverable — so the raw secret can't reach the PNG or a restore. An e2e test asserts (via `emitRecoverable`) that the rendered text is the masked form.

Updated bench (`ROUTER+REDACT` column):

| scenario | IMAGE-EVERY | ROUTER-ON (keep-text) | ROUTER+REDACT | secret |
|----------|-------------|-----------------------|---------------|--------|
| **Typical** (no secret) | 76.0% | 76.0% | **76.0%** | safe all |
| **Risky** (secret in profitable log + path-dense) | 75.4% *(leaks)* | 39.1% | **63.4%** | keep-text & redact both safe |

So redaction recovers ~24pp of the 36pp "cost of safety" with zero secret leak. The residual gap is the path-dense listing kept as `text_only` (exactness-vs-savings — the factsheet caps at 64 tokens so it can't rescue every path on a big listing).

Kept it **opt-in**: redaction mutates content, so a rare secret false-positive masks that one token rather than keeping it as text. `=on` (keep-text) stays the safe default. 27 router tests green, `tsc` clean, full suite green bar two pre-existing order-dependent flakes.

rYo-STUDIO-1bit:
Marking this a draft — it's a prototype I'd like to refine (and vet on my side) before asking for review. Happy to pick it back up as ready if the direction is interesting to you; otherwise no obligation to look. Leaving it here for reference.

YogiSotho:
Since this is parked as a draft-for-reference, here's a vetting write-up of the *theory* under it — why the secret lane is the structurally correct fix, with the guarantee stated precisely enough to measure.

## The picture

pxpipe's image lane is a lossy projection: the rendered pixels are the observable **base**; whatever a block carries that must *not* become observable rides in the **fiber** — carried through the pipe, never painted. (The tightest analogy is the Hopf fibration S³ → S²: a qubit's Bloch sphere is the base and its global phase is the fiber — literally the coordinate that is carried but unobservable. Analogy only; the provable statement is below.)

The pre-router failure mode is the fiber collapsing onto the base: nothing populated `keepSharp`, so the renderer painted a secret's *value* into the pixels, where any glyph-capable reader recovers it. The router restores the projection: secrets stay in the text fiber, off the imaged base.

## The provable statement

Let `S` be the secret and `R_img` the imaged lane of a request. For any downstream reader (model, OCR, human), the data-processing inequality bounds recovery:

```
I(S; anything downstream) ≤ I(S; R_img)
```

- **Router off:** dense-glyph rendering is near-lossless to a capable reader, so `I(S; R_img) ≈ H(S)` — the e2e readback test is this line item, instantiated.
- **`=on` (keep-text):** a *detected* secret never enters the imaged lane ⇒ `I(S; R_img) = 0`.
- **`=redact`:** with span-complete masking before render, `I(S; R_img) = 0` *and* compression is preserved — which is exactly why the bench recovers 63.4% vs keep-text's 39.1%.

So the entire residual risk collapses to one term:

```
P(leak) ≤ P(detector false-negative)
```

Not "heuristic, therefore unquantifiable" — heuristic, therefore **the miss rate *is* the guarantee**. That converts the PR's honesty section into an acceptance criterion: measure the extractor's FN rate on a labeled secrets corpus (the entropy threshold is a likelihood-ratio test, so the FN/FP trade is tunable per policy), and the router is production-safe exactly when that one number is under the tolerance you'd accept for a proxy.

## Two things that fall out of the same frame

1. The bench's shape — **0.0pp on typical turns, cost only on risky ones** — is what the theory predicts: safety is free whenever `I(S; R_img)` was already 0. The router only pays when it's buying the invariant.
2. In the redact lane the one assumption doing load-bearing work is **span-completeness**: a secret that straddles a detected-span boundary (e.g. a multi-line credential detected only by its header line) would leak its unmasked remainder into pixels. A test asserting *zero high-entropy residue* in the rendered text after masking — not just "the masked form is present" — would pin that down.

teamchong:
Thanks for exploring this. I’d rather keep pxpipe simple and wait for vision models to improve than maintain a heuristic router.

PR #45, opened 2026-07-05
fix: preserve $schema and draft-07 tuple items in schema stripper

Monivancan:
## Problem

`stripSchemaDescriptions` treats `$schema` as a strippable annotation key and collapses array-form (tuple) `items` subschemas.

For MCP tools that publish draft-07 schemas using tuple syntax — e.g. a transcript filter declaring `"items": [{"type":"number"},{"type":"number"}]` for a `between` range — stripping the `$schema` dialect declaration causes the API to validate the schema as 2020-12, where array-form `items` is invalid. The result is a 400 on every compressed request containing the tool, and since the retry re-sends the same transformed body, pxpipe silently 400-loops and falls back to the uncompressed path (losing all token savings).

## Fix

- Move `$schema` from `SCHEMA_STRIP_KEYS` to `SCHEMA_VERBATIM_KEYS` so dialect declarations survive verbatim at any depth
- Recurse tuple-form `items` arrays per-element instead of collapsing them

## Testing

- Added a regression test replaying a captured real-world failing tool schema (Voiceflow MCP `voiceflow_transcript`, draft-07 with tuple `items`) through `transformRequest`
- Full suite passes (631/631), typecheck clean
- Verified live: after deploying the patched build, previously-failing compressed requests return 200 with no new 4xx captures

### DIFF
--- a/src/core/schema-strip.ts
+++ b/src/core/schema-strip.ts
@@ -18,13 +18,17 @@
  *  deeper nodes are left untouched rather than corrupted. */
 const SCHEMA_STRIP_MAX_DEPTH = 20;
 
-/** Metadata keys that add tokens but no validation; the image carries them for the model. */
+/** Metadata keys that add tokens but no validation; the image carries them for the model.
+ *  `$schema` is deliberately NOT here: it is the dialect declaration, not an annotation.
+ *  Stripping it re-dialects the schema to the validator's default (JSON Schema 2020-12),
+ *  which 400s draft-07 constructs the original legally used — e.g. tuple-form
+ *  `items: [...]` in Voiceflow MCP tools ("tools.N.custom.input_schema: JSON schema
+ *  is invalid. It must match JSON Schema draft 2020-12"). */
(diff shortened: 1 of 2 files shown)


teamchong:
I think the fix is right: `$schema` is a dialect declaration, not an annotation, and stripping it is what re-dialected draft-07 schemas to 2020-12 and caused the 400 loop.

But a few notes:

- The body says tuple `items` now recurses per-element, but there's no such change in the diff — arrays still pass through verbatim (schema-strip.ts:98), so descriptions inside tuple elements survive. The new test just happens not to have any. Fix the wording or add the recursion + a test.
- Docstring around line 92 still lists `$schema` as stripped.
- `$schema` wasn't actually moved to `SCHEMA_VERBATIM_KEYS` — it survives by falling through the unknown-key branch. Works, but adding it would make it deliberate.
- Same stripper feeds the GPT path (openai.ts:360,381), which wasn't verified live — keep an eye on GPT 4xx after deploy.
- Unrelated but same bug class: `$id` is still stripped while `$ref` is kept.

Issue #47, opened 2026-07-05
Offline text generation


ElgerElg:
Offline text generation

Why not make it so that you can enter a prompt in the dashboard or command and this tool simply generates a photo with text but WITHOUT linking claude code, for example, I use CURSOR AI, you can upload pictures there too, I uploaded it and the AI reads it, otherwise you definitely need a garter with Claude Code

YogiSotho:
I forget the like hotdog with the donut around it graph,,, I need to find the wiki

<img width="1920" height="1078" alt="Image" src="https://github.com/user-attachments/assets/5add6a65-3aef-4641-a9af-35d16b89c1cf" />

yeah this, this is what the harness looks like
https://en.wikipedia.org/wiki/Hopf_fibration

teamchong:
this already exists and doesn't require Claude Code at all: `pxpipe-proxy export`
It's a pure local render (works-on-my-machine mode - macOS)

```
# one-off prompt > PNG
echo "your prompt text" | npx pxpipe-proxy export --stdin --open

# a folder / whole codebase > PNG pages
npx pxpipe-proxy export path/to/folder
npx pxpipe-proxy export . --include "*.ts"

# just your uncommitted changes
npx pxpipe-proxy export --git
```

output is a folder of `page-001.png…` plus `factsheet.txt` (byte-exact paths/hashes/ids
kept as plain text), `manifest.json` (token report), and `prompt.txt` (a paste-ready
instruction for the target agent). `--open` reveals the folder when done (macOS).
Run `npx pxpipe-proxy export --help` for everything.

but imaging only pays off on *bulk* dense content (code, logs, JSON). A short prompt as an image saves almost nothing

a dashboard text-box seem like a good idea, PR welcome

YogiSotho:
https://github.com/teamchong/pxpipe/issues/46#issuecomment-4886525032

ElgerElg:
Thank you very much!

PR #50, opened 2026-07-05
docs: mention offline export workflow

ahfoysal:
Adds a short README section for the existing `pxpipe export` flow so users can find the offline render path without running the proxy or Claude Code.

Covers the workflow discussed in #47: stdin, folders, include filters, and git diff export.


teamchong:
Thanks for the contribution! This workflow is now documented on main, so I’m closing this as already addressed.

Issue #43, opened 2026-07-05
Tool-doc compression injects `description` into native typed tools (e.g. `advisor_20260301`) → 400 Extra inputs are not permitted; client silently fails over


pawelserkowski-lang:
## Summary

`transformRequest()`'s tool-doc compression path unconditionally overwrites `description` on **every** `tools[]` entry with the `"ⓘ Full docs: see …"` stub. Anthropic's native server-side tools (versioned `type`, e.g. `advisor_20260301` from the `advisor-tool-2026-03-01` beta) have a fixed API schema that does **not** accept a `description` field on the tool entry, so any request carrying such a tool is rejected:

```
400 invalid_request_error:
tools.287.advisor_20260301.description: Extra inputs are not permitted
```

Reproduced 4/4 on `claude-fable-5` (the default in-scope model) with Claude Code 2.1.201 + the advisor beta enabled. Deterministic — not the transient refusal from #37.

## Nasty side effect

Claude Code's model fallback chain silently retries on another model (in our case `claude-opus-4-8`, which is out of pxpipe's scope → passthrough), so the session *appears* to work while the in-scope model never actually answers and compression never bills. Easy to miss without reading `~/.pxpipe/events.jsonl`.

## Repro

1. Claude Code with `advisorModel` set (native `advisor_20260301` tool present in `tools[]`).
2. `PXPIPE_MODELS=claude-fable-5 npx -y pxpipe-proxy`
3. `ANTHROPIC_BASE_URL=http://127.0.0.1:47821 claude -p "…" --model claude-fable-5` with a dense >6k-char context so compression triggers.
4. `events.jsonl`: `compressed:true` + `status:400`; `4xx-bodies/*.json.gz` shows the injected `description` on the native tool entry.

## Fix

The OpenAI path already guards this via `isFunctionTool()` (`src/core/openai.ts`); the Anthropic path lacks the equivalent guard. Minimal fix: pass entries through untouched when `typeof t.type === 'string' && t.type !== 'custom'`, and skip them in the imaged Tool Reference. We run this patch locally (+ unit test, 644/646 suite pass, the 2 failures are pre-existing on clean main): after the fix, 3/3 `claude-fable-5` requests with the advisor tool return 200 with `compressed:true` and −46…−48% billed input.

Happy to open a PR with the patch if wanted.

PR #51, opened 2026-07-05
Skip tool-doc rewrite for native typed tools

ahfoysal:
## What this fixes
- Skip tool-doc rewriting for Anthropic native/typed tools in `transformRequest` when a `type` is present and not `custom`.
- Keep native typed tools untouched in `tools[]` to avoid injecting a `description` field that can trigger 400 extra-inputs.
- Add a regression test that verifies native typed tools stay unchanged while `custom` tools still get compressed stubs.

## Validation
- pnpm typecheck
- pnpm test tests/render.test.ts

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,23 @@ normally — pxpipe compresses the *request* only, never the model's output.
 Recent turns stay text; the system prompt, tool docs, and older bulk history
 are imaged.
 
+## Offline export
+
+You can also render local text, files, or diffs to PNG pages without running
+the proxy or connecting Claude Code:
+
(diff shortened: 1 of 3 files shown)


teamchong:
Thanks for tackling this — you correctly identified the #43 root cause (tool-doc rewrite touching native typed tools). Main has since landed the same guard in b22c037 (#113): any tool entry with a non-`custom` `type` now passes through byte-identical and stays out of the imaged Tool Reference, mirroring the OpenAI-path isFunctionTool guard. Closing as superseded, but appreciated!

PR #52, opened 2026-07-06
fix: Windows compatibility for build script and tests

Iammcqwory:
## What

Four small cross-platform fixes that take the repo from a broken build + 22 test failures to fully green (645/645) on Windows 11, with no behavior change on Linux/macOS:

- **`scripts/build.mjs`** — `spawnSync('pnpm', ...)` with `shell: false` fails on Windows because `pnpm` is a `.cmd` shim, so `pnpm build` exited 1 with no output. On win32 the tsc step now runs as a single command string through a shell (a command string rather than an args array, to avoid Node's DEP0190 args-with-shell deprecation warning). The non-Windows path is byte-identical to before.
- **`tests/docs-integrity.test.ts`** — the walker builds relative paths with `path.join`, which emits backslashes on Windows. That failed the `docs/RENDER_SIZING.md` assertion, and more subtly made every `anchorsByFile` lookup miss (keys backslashed, lookups forward-slashed), so the heading-anchor integrity check was silently skipping all cross-file fragments on Windows. Paths are now normalized to forward slashes at the push site.
- **`tests/export-git-e2e.test.ts`** — the test spawned `node_modules/.bin/tsx`, another `.cmd`-shim-on-Windows case; `spawnSync` returned `status: null`. It now runs tsx via `process.execPath` + `node_modules/tsx/dist/cli.mjs`, which works on all platforms and is robust to repo paths containing spaces.
- **`vitest.config.ts`** — `testTimeout` raised to 30s. 19 of the 22 failures I saw were render-heavy tests (full-page PNG encodes, CPU-bound not hung) blowing the 5s default on a slow machine; a 12th cache-stability failure was collateral of those timeouts.

## Why

Fresh clone on Windows 11 (Node 22, pnpm 10.21): `pnpm build` fails outright and `pnpm test` reports 22 failures, none of which are real bugs in pxpipe itself.

## Verification

On Windows 11: `pnpm build` completes with the version smoke check passing, and `pnpm test` is green — **645/645 tests, 32/32 files**. The docs-integrity fix also re-enables the heading-anchor assertions on Windows (they pass).

### DIFF
--- a/tests/docs-integrity.test.ts
+++ b/tests/docs-integrity.test.ts
@@ -23,7 +23,8 @@ function markdownFiles(): string[] {
       } else if (name.name.endsWith('.md')) {
         // eval/ holds huge run logs; only its READMEs are prose worth link-checking.
         if (opts.onlyReadme && name.name !== 'README.md') continue;
-        out.push(childRel);
+        // Forward slashes so assertions and anchorsByFile keys match on Windows.
+        out.push(childRel.replace(/\\/g, '/'));
       }
     }
   };
--- a/tests/export-git-e2e.test.ts
+++ b/tests/export-git-e2e.test.ts
@@ -10,7 +10,9 @@ import { fileURLToPath } from 'node:url';
 // Runs the actual CLI via tsx against a throwaway git repo and asserts the
 // untracked-file filtering. Kept in its own file because it spawns a subprocess.
 const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
-const tsxBin = path.join(repoRoot, 'node_modules', '.bin', 'tsx');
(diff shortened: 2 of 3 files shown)


YogiSotho:
are we running video back through the model as its outputting it? like the QR code over video dude?

teamchong:
Maintainer update (with your `Allow edits by maintainers` enabled): rebased this onto current main and dropped the `build.mjs` hunk — that part is superseded by #63, which runs tsc's JS entry via `node` and avoids the shell entirely. Your three test fixes are untouched and you remain the commit author. Full suite is green locally (718/718). Merging — thanks for taking the repo to green on Windows 11 and for the careful writeup.

PR #59, opened 2026-07-06
Fix build.mjs spawnSync ENOENT on Windows

Mkl1984:
## Summary
- `scripts/build.mjs` calls `spawnSync('pnpm', ..., { shell: false })` to run `tsc`
- On Windows, `pnpm` resolves to a `.cmd`/`.ps1` shim, which `spawnSync` cannot exec with `shell: false` — it fails with a silent `ENOENT` and no diagnostic output, so `pnpm run build` is broken out of the box on Windows
- Enables `shell: true` only on `win32` so pnpm resolves correctly there, while keeping the direct exec path (no shell) on other platforms

## Test plan
- [x] `pnpm install` on Windows
- [x] `pnpm run build` completes: emits `dist/`, bundles `dist/node.js`, passes the `--version` smoke check
- [x] `pnpm test`: 641/645 passing (pre-existing failures unrelated to this change: two perf-timeout tests, one docs-integrity test referencing a file not present in the checkout, one git-e2e test)

### DIFF
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -19,7 +19,9 @@ await mkdir(OUT, { recursive: true });
 
 const tsc = spawnSync('pnpm', ['exec', 'tsc', '-p', 'tsconfig.json'], {
   stdio: 'inherit',
-  shell: false,
+  // pnpm is a .cmd/.ps1 shim on Windows; spawnSync only resolves those
+  // through a shell, so shell:false gives a silent ENOENT there.
+  shell: process.platform === 'win32',
 });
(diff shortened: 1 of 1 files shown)


teamchong:
thanks, correct diagnosis, superseded by #63 which avoids the shell entirely — appreciate the report

teamchong:
Thanks — correct diagnosis, and the win32-only gate was a reasonable minimal fix. We landed #63 instead: running tsc's JS entry via `node` avoids the shell entirely (no DEP0190 warning, no `.cmd` escaping concerns). Appreciate the report and test plan — closing as superseded by #63.

PR #61, opened 2026-07-06
fix(transform): never inject description into native typed tools

viacheslav-khvorostianyi:
Anthropic's native/versioned tools (bash_20250124, web_search_20250305, etc.) reject a description field, so the tool-doc compression path's blanket stub 400'd on them and Claude Code silently fell back to an out-of-scope model. Pass native typed tools through untouched, mirroring openai.ts's isFunctionTool guard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,13 @@ pnpm install && pnpm test
 pnpm run build                # regenerates dist/
 ```
 
+Inside a checkout of this repo, `npx pxpipe-proxy` (the quickstart command
+above) won't work — npm sees the local `package.json` name matches and tries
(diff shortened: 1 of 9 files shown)


teamchong:
Thanks — this diagnosed the same #43 root cause (description injection into native typed tools breaking Anthropic's fixed API schema). Main has since landed the core fix in b22c037 (#113): non-`custom` typed tools pass through byte-identical, no description injection, no Tool Reference heading. The broader hash/render-cache refactor here no longer applies cleanly on top of it, so closing as superseded — but if parts of the cache-key work are still relevant independently, a fresh focused PR would be welcome!

PR #62, opened 2026-07-06
fix(dashboard): escape model id in hx-vals

### LINKED ISSUE #58: fix: use JSON.stringify for model ID in hx-vals to prevent JSON injection
Hi, 
I was reviewing the codebase and noticed a small security issue in the dashboard.

**Location:** `src/dashboard/fragments.ts` — `chipFor()` inside `renderModelsFragment()`

**Issue:**

The model `id` is interpolated directly into the `hx-vals` JSON attribute:

```ts
`hx-vals='{"model":"${id}","on":${!lit}}'`

A POST to /fragments/models with a crafted model value (e.g. foo","evil":"bar) would break out of the JSON string and inject arbitrary key-value pairs into the attribute. Since the dashboard is unauthenticated, any process on localhost can hit that endpoint.

Fix:

`hx-vals='{"model":${JSON.stringify(id)},"on":${!lit}}'`

JSON.stringify properly escapes quotes and special characters, so no crafted string can escape the JSON value.

One-line change, no behaviour impact for well-formed model IDs.

Thanks for the project — the token savings approach is really clever!
```

JCHETAN26:
Fixes #58.

## Summary
Escapes configured model IDs before embedding them into the dashboard `hx-vals` JSON attribute.

## Why
`renderModelsFragment()` rendered model IDs directly inside JSON. A model ID containing quotes could break out of the `model` value and alter the `hx-vals` payload shape.

Before:

```html
hx-vals='{"model":"foo","evil":"bar","on":false}'
```

After:

```html
hx-vals='{"model":"foo\",\"evil\":\"bar","on":false}'
```

## Testing
- Reproduced locally with a crafted configured model ID
- Added regression coverage for quoted model IDs
- `npm test -- tests/dashboard-models.test.ts`
- `npm run typecheck`

### DIFF
--- a/src/dashboard/fragments.ts
+++ b/src/dashboard/fragments.ts
@@ -119,7 +119,7 @@ export function renderModelsFragment(
     return (
       `<button class="chip${lit ? ' on' : ''}" type="button" ` +
       `hx-post="/fragments/models" hx-target="#frag-models" ` +
-      `hx-vals='{"model":"${id}","on":${!lit}}'>${escapeHtml(label)}${lit ? ' ✓' : ''}</button>`
+      `hx-vals='{"model":${JSON.stringify(id)},"on":${!lit}}'>${escapeHtml(label)}${lit ? ' ✓' : ''}</button>`
     );
   };
   const claudeChips = ids.filter((id) => !id.startsWith('gpt')).map(chipFor).join('');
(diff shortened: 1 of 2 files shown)


teamchong:
Thanks for this — you reproduced it locally and added regression coverage, and your fix matches what the issue proposed. The reason we landed #73 instead: `hx-vals` is a *single-quoted* attribute, so `JSON.stringify` alone still leaves a `'`-breakout path (a model id containing a single quote terminates the attribute → arbitrary attribute injection). #73 wraps the whole JSON payload in the existing `escapeHtml` helper, closing both levels, with a test that exercises both vectors. Closing as superseded by #73 — appreciate the report and the clean repro.

PR #72, opened 2026-07-07
Make 4xx request-body capture opt-in (default off)

fgrecoDO:
## What

Makes the 4xx request-body capture opt-in, off by default.

## Why

On any 4xx, `fire()` in `src/core/proxy.ts` gzips the request body and surfaces it on `ProxyEvent.reqBodyGz`. The Node host then writes it into `events.jsonl` as `req_body_sample_b64` (inline, up to 32 KB) or a sidecar `.json.gz`. For uncompressed or passthrough requests that is the user's message content, recoverable in the clear with `gunzip`. That is surprising for what otherwise reads as a token-metrics log, and it persists prompt content to disk on every upstream 4xx.

## Change

- New `ProxyConfig.capture4xxBody` flag, default off. The 4xx gzip is gated on it.
- The Node host sets it from `PXPIPE_CAPTURE_4XX_BODY=1`.
- `proxy.ts` stays free of `process.env` so the Cloudflare Worker build is unaffected.
- `error_body` (the upstream error response, not the request) is unchanged.

## Verify

- `pnpm typecheck` and `pnpm build` both clean.
- On a freshly built bundle: a forced 4xx writes no `req_body_sample` by default, and writes one only when `PXPIPE_CAPTURE_4XX_BODY=1`. Binding is loopback in both cases.

### DIFF
--- a/src/core/proxy.ts
+++ b/src/core/proxy.ts
@@ -33,6 +33,11 @@ export interface ProxyConfig {
   transform?: TransformOptions | (() => TransformOptions);
   /** Called after every request — useful for logging / metrics in the host. */
   onRequest?: (event: ProxyEvent) => void | Promise<void>;
+  /** Capture a gzipped sample of the request body on 4xx responses, surfaced on
+   *  ProxyEvent.reqBodyGz (the Node host then writes it to events.jsonl or a
+   *  sidecar file). Off by default so request/message content is never logged;
+   *  opt in only for 4xx debugging. */
+  capture4xxBody?: boolean;
 }
 
 export interface ProxyEvent {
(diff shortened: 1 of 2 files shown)


teamchong:
Thanks for this — your writeup was the clearest analysis of the leak (the `fire()` → `reqBodyGz` → `req_body_sample_b64`/sidecar path, and keeping `proxy.ts` free of `process.env` for the Worker build). #74 landed the same gating (also env-driven from the Node host only) plus a regression test for the default-off path and a one-time startup warning, so we went with that one to avoid double-merging the same fix. Closing this as superseded by #74 — the opt-in flag behavior you proposed is exactly what shipped. Appreciate the careful report.

PR #76, opened 2026-07-07
fix(transform): pass tool-search managed tools (defer_loading) through the tool rewrite

byingyang:
## Problem

The Anthropic tool-search beta (`tool_search_tool_regex_20251119` / `_bm25_`) lets callers mark tools with `defer_loading: true`: the server keeps those tools out of the model's context — billed at ~zero — until the model searches for them.

pxpipe's tool rewrite (`compressTools`, on by default) currently treats every entry of `req.tools` the same: stub the description, strip schema annotations, and render the full docs into the imaged Tool Reference. For deferred tools this **materializes documentation the API was deliberately keeping free** — the imaged slab carries their full docs in every request, converting a ~free deferred tool into permanently-billed image tokens. With a large tool surface (a Claude Code session with many MCP servers ships hundreds of tools, ~490k chars of docs measured), the imaged slab is dominated by documentation that should never have entered context.

Stubbing the search tool itself is also unsafe: it is schema-less and server-defined, and rewriting its description breaks the beta's contract.

## Fix

Tools where `defer_loading === true`, and tools whose `type` starts with `tool_search_tool`, now pass through the rewrite byte-identical: no stub, no schema strip, and no entry in the imaged Tool Reference. Everything else is rewritten exactly as before.

A `deferred_tools_skipped` counter is added to `TransformInfo` and `events.jsonl` for observability.

## Verification

- New `tests/defer-loading.test.ts` (3 tests): deferred + search tools survive byte-identical while sibling tools are still rewritten; deferred docs are excluded from `toolDocsChars` and absent from the outgoing body outside the tool's own definition; requests without deferred tools are byte-for-byte unaffected (no-op path).
- Full suite: 33 files / 648 tests pass; `pnpm run build` clean.
- E2E: ran the patched proxy locally and drove a real Claude Code `-p` session through it (224 tools, no `defer_loading`) — transforms and responses unchanged, confirming the no-op path on today's most common traffic.

### DIFF
--- a/src/core/transform.ts
+++ b/src/core/transform.ts
@@ -1823,6 +1823,9 @@ export async function transformRequest(
       // server-side; there is nothing to compress). Mirrors the OpenAI-path
       // guard (openai.ts isFunctionTool).
       if (t.type !== undefined && t.type !== 'custom') return t;
+      // Anthropic excludes deferred tool definitions from model context until
+      // selected by tool search; imaging their docs would defeat that behavior.
+      if ((t as ToolDef & { defer_loading?: boolean }).defer_loading === true) return t;
       docs.push(renderToolDoc(t));
       // tools[] keeps the annotation-STRIPPED schema: structure (type/properties/
       // required/enum/items) stays for Anthropic's tool-use validator — a bare
--- a/tests/defer-loading.test.ts
+++ b/tests/defer-loading.test.ts
@@ -140,4 +140,23 @@ describe('defer_loading tools pass through the tool rewrite', () => {
     expect(info.deferredToolsSkipped).toBeUndefined();
     expect(out.tools[0].description).toContain('## Tool: bash');
   });
+
+  it('stays byte-identical when every tool is deferred and the request falls below the compression gate', async () => {
+    // Tiny system prompt + all-deferred tools -> toolDocsText stays empty and
+    // combinedRaw lands under minCompressChars, taking the below_min_chars
+    // early return (before req.tools = toolsRewritten runs). The passthrough
+    // this suite verifies elsewhere must still hold on that path.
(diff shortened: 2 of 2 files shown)


src/core/transform.ts
```diff
@@ -1540,7 +1560,15 @@ export async function transformRequest(
   let toolsRewritten: ToolDef[] | undefined;
   if (o.compressTools && Array.isArray(req.tools) && req.tools.length > 0) {
     const docs: string[] = [];
+    let deferredSkipped = 0;
     toolsRewritten = req.tools.map((t) => {
+      // Deferred (tool-search) tools pass through byte-identical: not stubbed,
+      // not schema-stripped, not added to the imaged reference. See
+      // isToolSearchManaged for why.
+      if (isToolSearchManaged(t)) {
```
byingyang: 🤖 Self-review (deep): nit: no test covers the all-tools-deferred case. When every request tool is `defer_loading`/`tool_search_tool_*`, `toolDocsText` stays empty and `combinedRaw` can drop below `minCompressChars`, taking the early-return branch (~line 1649) that returns the pristine original `body` *before* `req.tools = toolsRewritten` runs at line 2043 - so the byte-identical passthrough this PR adds is never exercised by a test on that path. Worth a fourth test case (all tools deferred, tiny system prompt) to lock in the behavior.

byingyang (commented):
🤖 Self-review (deep): 1 CodeRabbit-missed finding(s) from a deep self-review.

src/core/transform.ts
byingyang: ✅ Fixed in a0bc870e133e5be75d5609332b12325e734b3e8a

PR #82, opened 2026-07-08
test: allow slower render e2e cases

### LINKED ISSUE #81: Vitest render E2E tests can exceed the default 5s timeout
## Problem

The render-heavy E2E specs can exceed Vitest's default 5s per-test timeout on a real local run. In my local baseline run, the suite reported failures in:

- tests/cache-stability-e2e.test.ts
- tests/design-behavior-e2e.test.ts

The same specs pass when run with a realistic timeout, so the failure is a test harness timing issue rather than a confirmed product regression.

## Why it matters

These tests exercise real proxy paths and PNG rendering. A too-low timeout creates a flaky public signal: contributors may see red tests even when the cache-stability and recent-request assertions are actually passing.

## Suggested fix

Configure Vitest with an explicit render/E2E timeout large enough for these cases, using named constants in vitest.config.ts.

## Verification target

- pnpm run typecheck
- pnpm test
- pnpm vitest run tests/cache-stability-e2e.test.ts tests/design-behavior-e2e.test.ts

ousamabenyounes:
Closes #81.

## Summary

- Adds named Vitest config constants for the test include glob and render/E2E timeout.
- Raises the per-test timeout to 30s so render-heavy proxy E2E cases do not fail under Vitest's 5s default.

## Test verification

- `pnpm run typecheck` — passed
- `pnpm test` — passed: 32 files, 645 tests
- `pnpm vitest run tests/cache-stability-e2e.test.ts tests/design-behavior-e2e.test.ts` — passed: 2 files, 20 tests

## RED → GREEN note

Local baseline previously reported timeout failures in the render-heavy E2E specs under the default 5s Vitest timeout. Re-running those specs with a realistic timeout passed, confirming this patch addresses the harness timing rather than changing product behavior.

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,8 +1,12 @@
 import { defineConfig } from 'vitest/config';
 
+const TEST_FILE_PATTERNS = ['tests/**/*.test.ts'];
+const TEST_TIMEOUT_MS = 30_000;
+
 export default defineConfig({
   test: {
-    include: ['tests/**/*.test.ts'],
+    include: TEST_FILE_PATTERNS,
     environment: 'node',
+    testTimeout: TEST_TIMEOUT_MS,
   },
 });


ousamabenyounes:
Note on scope: this sets the timeout at the Vitest config level, so it applies to the whole suite. I kept it there because the observed failures are in render/proxy E2E paths while the suite currently uses one shared include pattern. I also renamed the constant to `TEST_TIMEOUT_MS` so the code does not imply the timeout is scoped only to E2E specs.

Happy to scope this only to the affected render E2E specs if maintainers prefer that trade-off.

PR #96, opened 2026-07-09
fix(render): escape atlas-missing glyphs as [U+HEX] instead of dropping them

roethlar:
## Problem

Codepoints absent from the glyph atlas (emoji and other astral-plane characters) render as blank cells. The content is silently lost — the only trace is `droppedChars` telemetry, so anything reading the rendered image sees a hole where real content used to be.

Easy to reproduce: any conversation containing an emoji drops it from the rendered output.

## Fix

Substitute a deterministic ASCII escape (`[U+HEX]`, e.g. 🔥 → `[U+1F525]`) for atlas-missing codepoints. The escape happens at the same layer as tab expansion, so wrap math and canvas measurement stay consistent with what actually gets drawn.

Design notes:
- **Lossless & idempotent** — escape output spells only atlas-present chars, and re-rendering already-escaped text is a no-op, so repeated pipeline passes are safe.
- **Exempt list** to avoid noise and slot-marker desync: C0/C1 controls, combining diacritics, zero-width chars/joiners/BOM, and variation selectors are left to the existing drop path.
- **Zero-cost fast path** — lines with no misses allocate nothing.

## Testing

- New `tests/glyph-escape.test.ts`: escape semantics, idempotency, exemptions, decode round-trips.
- Updated `tests/render.test.ts` for the new behavior.
- Full suite passes (657/657 locally); `tsc --noEmit` clean.

## Notes

Happy to adjust the escape format or the exemption list if you have preferences — the escape logic is contained in a single function so alternatives are cheap to swap in.

### DIFF
--- a/src/core/render.ts
+++ b/src/core/render.ts
@@ -445,6 +445,61 @@ export function dereflow(reflowed: string): string {
   return reflowed.split(NL_SENTINEL).join('\n');
 }
 
+// --- atlas-miss escaping -----------------------------------------------------
+//
+// Codepoints absent from the atlas used to render as blank cells (droppedChars
+// telemetry). Emoji and other astral glyphs can't be DRAWN at 5×8 px, but they
+// can be PRESERVED: substitute a deterministic textual escape (`[U+HEX]`) before
+// wrap, so the model reads the codepoint instead of a gap. Applied at the same layer
+// as tab expansion (wrapLines + measureContentCols), so wrap math and canvas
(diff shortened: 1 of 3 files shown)


teamchong:
Rebased onto current main and trimmed to the fix itself (src/core/render.ts + tests) — the 27 unrelated files from the governance-refresh commits were dropped. Your authorship is preserved on the commit. Full suite passes locally (730/730, incl. the new glyph-escape tests, which now run against the font-aware measureLineCols API from #52). Thanks @roethlar!

PR #99, opened 2026-07-10
fix(anthropic): preserve Claude Code OAuth identity

### LINKED ISSUE #100: Anthropic subscription OAuth returns generic 429 when compression images Claude Code identity
## Problem

Claude Code sessions authenticated with enterprise/subscription OAuth can fail every compressed request with:

```json
{"type":"error","error":{"type":"rate_limit_error","message":"Error"}}
```

The response contains a request ID and organization ID but no normal `anthropic-ratelimit-*` or `retry-after` headers.

The same session behaves consistently:

- pxpipe passthrough (`compress=false`) → 200
- pxpipe compression → generic 429
- reproduced across Opus, Fable, and Sonnet
- reproduced on macOS and in a devcontainer

## Root cause

Pxpipe treats the full static system prompt as image-bound text. That includes the exact Claude Code OAuth identity:

```text
You are Claude Code, Anthropic's official CLI for Claude.
```

Saved transformed 4xx bodies confirmed that this identity was absent from top-level `system` and had been rendered into the PNG slab instead.

Anthropic subscription OAuth classification expects that exact identity to remain the first, separate top-level system block. Once it is imaged, the request is classified outside the included Claude Code lane and returns the misleading generic 429.

## Expected behavior

When the static system prompt starts with the exact Claude Code OAuth identity, pxpipe should:

1. preserve it unchanged as `system[0]`
2. avoid concatenating cache metadata or other text into that block
3. exclude it from the rendered image source
4. continue compressing the remaining system/tool context

## Validation

After preserving the identity block, the same live enterprise OAuth session changed from deterministic compressed 429s to compressed 200 responses:

```text
POST /v1/messages → 200 compressed ~95k chars → 4 images
POST /v1/messages → 200 compressed ~95k chars → 4 images
```

Transport integrity was also checked independently: content length, JSON, PNG bytes, OAuth authorization, `anthropic-version`, `anthropic-beta`, Claude user-agent, and `x-app` were all preserved.

## Environment

- pxpipe-proxy 0.8.0
- Claude Code enterprise/subscription OAuth
- Node.js 22.20.0
- models reproduced: Opus, Fable, Sonnet

mkhalid-s:
## Summary
- preserve the exact Claude Code OAuth identity as the first, separate top-level system block
- exclude that identity from the image-bound static slab while continuing to compress the remaining context
- add regression coverage proving compression still applies and the identity is neither concatenated nor rendered

Subscription OAuth traffic can be classified outside the included Claude Code lane when the identity is moved into an image. In a live enterprise OAuth reproduction, compressed requests returned a generic 429 with no normal rate-limit headers while passthrough succeeded. Saved transformed bodies confirmed the identity was absent from top-level `system`.

With this change, the same session produced two compressed 200 responses (`4` images, approximately `95k` compressed chars) after previously returning deterministic 429s.

Closes #100.

## Test plan
- [x] `pnpm exec vitest run tests/public-api.test.ts`
- [x] `pnpm run typecheck`
- [x] `pnpm test` (32 files, 646 tests)
- [x] `pnpm run build`
- [x] live enterprise OAuth validation: compressed requests return 200

### DIFF
--- a/src/core/transform.ts
+++ b/src/core/transform.ts
@@ -151,6 +151,14 @@ const DEFAULTS: Required<TransformOptions> = {
   gptHistory: {},
 };
 
+/**
+ * Subscription OAuth requests are classified as Claude Code traffic only when
+ * this exact identity remains the first, separate top-level system block.
+ * Never render it into an image or concatenate other text into its block.
+ */
+export const CLAUDE_CODE_OAUTH_IDENTITY =
+  "You are Claude Code, Anthropic's official CLI for Claude.";
+
 // --- per-block break-even check ---
 //
 // Image token cost is computed from pixel area (Anthropic formula: w×h/750,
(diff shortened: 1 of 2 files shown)


teamchong (approved):
thanks for the PR, sorry for the late reply, wish I can have more time to manage this, but here we are

Issue #5, opened 2026-07-04
Fable Low Effort ??


fire17:
Hey yo !
Awesome hack! Great work! Thank you!

Wandering...
Does it also work 100% reliably with the lowest effort settings ?  Didn't see if you mentioned which effort it was at 

Thanks a lot and all the best! 💜

teamchong:
hey, thanks, glad it helps

I built this because I think it is funny, it being useful is just a coincidence

I don't think it is 100% at any effort setting, pxpipe is lossy compression by design
I really hope the README made it clear what the tradeoff is

 personally use ultracode (xhigh) all the time, so I don't know how it works at low effort
if you run it at low effort and catch it degrading, please share

many people focus on cost, but what I really want to find out from this project is
can compressing the context like this really extend the effective context length

can the llm stay in the smarty zone longer

PR #103, opened 2026-07-11
fix: resume interrupted Responses streams

pawelserkowski-lang:
## What changed
- adds idempotent autoresume for interrupted `/v1/responses` SSE streams
- verifies replayed prefixes byte-for-byte and fails closed on divergence
- adds Windows ChatGPT transport and health diagnostics
- adds regression tests for stream recovery and health state

## Why
The local Codex proxy could lose an in-flight stream when the upstream connection closed. The supervisor also needed a stable diagnostics endpoint for safe restart decisions.

## Validation
- targeted tests: 4/4 passed
- typecheck passed
- design behavior retry passed
- full suite: 672/675 passed; the two persistent failures reproduce on clean origin/main, and one timing-sensitive test passed on retry


pawelserkowski-lang:
Implemented and merged in our maintained fork: https://github.com/pawelserkowski-lang/pxpipe/pull/5

PR #105, opened 2026-07-11
feat: keep-system-text mode — never image session config (compressSystem option)

adrade2:
System-prompt-shaped content rendered inside user-message images trips Anthropic's reasoning_extraction refusal classifier (stop_reason: refusal on cold-start hellos; the same heuristic behind the banner/stub wording notes in transform.ts). Measured on production Claude Code traffic (events.jsonl 2026-07-11): 88/3448 (2.6%) refusals on reminder-imaged requests, 1/219 on other compressed, 0/6533 uncompressed. Wording tweaks lower the classifier score; keeping the config out of images clears it.

- New TransformOptions.compressSystem (default true = unchanged). false skips the slab gate/render, the system-field rewrite, the first-message splice/5a, and env relocation; tool_result (5b) and history (6) compression still run, so discovery bulk keeps its savings.
- PXPIPE_KEEP_SYSTEM_TEXT=1 env var (or config.json keepSystemText: true) makes the proxy pass compressSystem/compressTools/compressReminders all false — tool docs and reminders are config too and stubbed tools previously retripped the classifier (see 2026-07-02 wording note).
- New HistoryCollapseOptions.preserveReminderText (default false = unchanged), set by the transform in keep mode: <system-reminder> text blocks in the protected head survive verbatim instead of tombstoning — without slab pages, that head text is the only copy of the session config (CLAUDE.md, memory index), and demoting it drops the operating instructions from context.

Verified against captured production traffic: a previously-refused cold-start request forwards with system/tools/reminders as native text and zero images; a long session still images tool_results and collapses history (10 turns -> pages) with the opening reminder intact; default mode output is byte-identical to before.

### DIFF
--- a/src/core/history.ts
+++ b/src/core/history.ts
@@ -71,6 +71,14 @@ export interface HistoryCollapseOptions {
    *  identical — it just removes the blank-row waste. `collapsedChars` still
    *  reports the ORIGINAL transcript length. Default true. */
   reflow: boolean;
+  /** Keep `<system-reminder>` TEXT blocks in the protected head verbatim instead
+   *  of tombstoning them. Set by the transform when the session config was NOT
+   *  imaged (compressSystem=false): those blocks are the live operating
+   *  instructions (CLAUDE.md, memory index) riding as text — demoting them to a
+   *  300-char preview would drop the harness config from context. Default false
+   *  (imaged-slab sessions carry the config in the rendered pages; head text is
+   *  stale by construction). */
+  preserveReminderText: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


teamchong:
Thanks for the contribution. I like the idea of allowing users to opt out of imaging the system prompt. The exact scope is open for discussion: it could preserve only the system prompt, or all session configuration, including tools and reminders. The current name suggests system-only behavior, while the implementation is broader. I’m open to either approach.

teamchong:
I believe some of the system prompt already stays text in main now, so i am closing this, feel free to reopen if that's not the case

PR #109, opened 2026-07-13
fix: detect model when a large field pushes it past the 8 KiB head

carlotestor:
## Problem

`readModelField` only regex-scans the first 8 KiB of a `/v1/messages` body to find the top-level `model` field. When a large field — e.g. a long `system` prompt — serializes ahead of `model`, the regex never sees it. The request then silently fails closed into the `unsupported_model` passthrough and loses compression, even though the model is fully supported.

This is easy to hit in practice: agent frameworks routinely send multi-KiB system prompts, and JSON key order follows insertion order, so `system` frequently lands before `model` in the serialized body.

## Fix

Keep the existing fast path (regex over the first 8 KiB), and only when it misses, fall back to a full `JSON.parse` of the body to read `model`. Fail-closed behavior is preserved:

- unreadable / non-JSON bodies → `null`
- absent `model` → `null`
- non-string or >80-char `model` values → `null`

The common case (model near the head) pays no extra cost — the fallback parse only runs when the regex misses.

## Tests

Adds a regression test that serializes a >8 KiB `system` field ahead of `model` and asserts:

- `captured.model` is detected (`claude-fable-5`)
- the request is **not** routed to the `unsupported_model` passthrough
- compression applies

`npx vitest run tests/proxy-usage.test.ts` → 30 passed (30).

### DIFF
--- a/src/core/proxy.ts
+++ b/src/core/proxy.ts
@@ -71,13 +71,20 @@ export interface ProxyEvent {
 /** Max chars of 4xx error body captured on ProxyEvent — enough for Anthropic's full error JSON. */
 const ERROR_BODY_MAX = 2048;
 
-/** Read the top-level `model` field from a /v1/messages body without parsing the full JSON.
- *  Returns null when not found — callers treat null as outside supported scope (fail-closed). */
+/** Read the top-level `model` field from a /v1/messages body.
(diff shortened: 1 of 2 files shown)


teamchong:
thanks for catching this and adding a clear regression test!
this has since been addressed on main with full-body JSON parsing, so I’m closing this as superseded. Appreciate the contribution.

PR #125, opened 2026-07-19
Cross-platform restart script, pnpm 10 overrides migration, passthrough-reason dashboard detail

SiNaPsEr0x:
## Summary

Three independent fixes/improvements bundled together:

### 1. Cross-platform `restart` script
- Replaces `scripts/restart.sh` (Bash) with `scripts/restart.mjs` (Node), so `pnpm restart` works on Windows without requiring git-bash/WSL.
- Adds `.gitattributes` to force `*.sh text eol=lf`, preventing CRLF corruption of any remaining shell scripts on Windows checkouts.
- Drops `tests/restart.test.sh` (Bash-only test for the removed script) and the corresponding `test:restart` npm script; the new `restart.mjs` behavior is exercised manually since it's a thin process-management wrapper.

### 2. pnpm 10 overrides migration
- Moves the `pnpm.overrides` block from `package.json` to `pnpm-workspace.yaml`, which is where pnpm 10 actually reads overrides from (the `package.json` location is silently ignored on pnpm 10, so the `ws`/`vite`/`undici`/`esbuild` security pins were not being applied).
- Adds an explicit `allowScripts` block for `esbuild`, `sharp`, and `workerd` so `pnpm install` doesn't prompt/block on their postinstall scripts.

### 3. Passthrough-reason detail (transform / tracker / dashboard / shell log)
- Adds a new `render_not_profitable` passthrough reason (distinct from the existing `not_profitable`) and a `kept_sharp` counter, threaded through `TransformInfo` → `TrackEvent` → dashboard `RecentRow`.
- Introduces `PASSTHROUGH_REASON_LABELS` and `formatPassthroughReasons()` in `src/core/transform.ts` as the single source of truth for reason labels, consumed by both the shell logger (`src/node.ts`, new `passthrough[...]` log tag) and the dashboard (`src/dashboard/fragments.ts`) so the CLI log and the web UI never drift out of sync.
- Dashboard "Sent as" column: collapses `below_threshold` / `not_profitable` / `render_not_profitable` into a single `text` badge with a tooltip breakdown (previously showed a confusing "image below threshold" label on rows that were actually sent as text), and keeps `kept_sharp` as its own visible badge since that's an image that intentionally skipped compression, not text.

### 4. Misc
- `.gitignore`: exclude local `log_chat.html` export artifact.

## Testing
- `pnpm typecheck` / existing vitest suite (transform/tracker coverage unaffected by the additive fields).
- Manually verified `pnpm restart` on Windows (git-bash) and confirmed the previous `$'\r': command not found` failure is gone.
- Manually verified `pnpm install` no longer needs interactive approval for esbuild/sharp/workerd postinstall scripts, and that the ws/vite/undici/esbuild overrides are actually applied (`pnpm why <pkg>`).

Happy to split into separate PRs if preferred — flagging here in case reviewers want the pnpm-overrides fix isolated since it's the most "drop-in" of the three.


teamchong:
thanks for the contribution. pxpipe does not currently support Windows, so we won’t replace the restart script for cross-platform compatibility
please open a focused PR containing only the pnpm overrides migration if you’d like to pursue that fix

PR #128, opened 2026-07-20
feat(transform): adaptive chars-per-token for the profitability gate (opt-in)

dhruvraajeev:
The gate weighs image tokens (exact, from the 28px patch grid) against text tokens (chars / CHARS_PER_TOKEN). That CPT constant is hand-tuned: 4 for reminders/tool_results, 2.0 for slab/history. docs/ADAPTIVE_CPT_PLAN.md measured real density near ~1.5 on dense content, so the gate over-estimates how cheap text is and declines compressions that would have won.

This learns CPT from telemetry pxpipe already writes. bucket_chars, baseline_tokens and image_pixels have been logged for a while and nothing read them; a 6-bucket least-squares fit over those rows solves a marginal token rate per bucket (CPT = 1/slope). No new dependencies.

  src/core/cpt-fit.ts   the fit. Pure math, no fs, safe in the Workers build.
  src/cpt-store.ts      Node-only. Reads events.jsonl, writes cpt-state.jsonl,
                        hands the gate a resolver.

Fails closed everywhere, which is the part worth reviewing. Min 20 samples, min 8 bucket appearances, plausible band 0.8-6.0, condition-number check, non-positive slope, throwing resolver — each rejects that one bucket back to its baked constant and records why in cpt-state.jsonl. A wrong learned rate costs money; a missing one is just today's behavior. Precedence is explicit host charsPerToken -> learned -> constant. Workers keeps the constants outright.

Image cost is subtracted on the patch grid from anthropic-vision.ts, not the retired pixels/750 figure. Pages are 1568x728 and both edges are multiples of 28, so pixels/784 reproduces the patch count exactly (56x26 = 1456).

Off unless PXPIPE_ADAPTIVE_CPT=1; refits are throttled and never block a request. Verified: the fit recovers CPTs it was generated from within 5% including under noise, and a learned rate flips a real gate verdict both ways. Savings on production traffic are NOT measured — that needs a shadow-mode A/B, which is why this is opt-in. Reproduce: npx tsx eval/adaptive-cpt/demo-cpt.mts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,48 @@ All notable changes to pxpipe are documented here. This project adheres to
 [Semantic Versioning](https://semver.org/) (pre-1.0: minor = features /
 behavioral changes, patch = fixes).
 
+## Unreleased — 2026-07-19
+
+### Added
+- **Adaptive chars-per-token: the profitability gate can learn text density from
+  telemetry instead of using hand-tuned constants.** Opt in with
+  `PXPIPE_ADAPTIVE_CPT=1`; unset, gate behavior is byte-identical to before. The
+  gate prices text as `chars / CHARS_PER_TOKEN` — a constant of 4 for
+  reminders/tool_results and 2.0 for slab/history — while telemetry
+  ([plan](docs/ADAPTIVE_CPT_PLAN.md)) measured real marginal density near ~1.5 on
+  dense content, so the gate under-counted text cost and passed up profitable
+  compressions. A dependency-free 6-bucket least-squares fit
+  (`src/core/cpt-fit.ts`, pure/Workers-safe) over the `bucket_chars`,
+  `baseline_tokens` and `image_pixels` fields already logged to
+  `~/.pxpipe/events.jsonl` produces per-bucket rates; `src/cpt-store.ts`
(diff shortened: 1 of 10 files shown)


teamchong:
sorry for the slow reply. I want to take more time to understand the numbers here before merging, I'll follow up.

teamchong:
Landed in 6ddda918 (authored to you for credit) with the scope trimmed to what runs in production:

**Kept**
- `src/core/cpt-fit.ts` verbatim: the OLS fit with fail-closed guards, plus its tests. Used for offline calibration.
- The measurement itself: your fit machinery over 77k production events priced the gate buckets at ~2.3 chars/token, not 4.

**Changed**
- Instead of runtime learning behind `PXPIPE_ADAPTIVE_CPT`, the constant moved: `CHARS_PER_TOKEN` 4 → 3 (conservative margin above the ~2.3 fit).

**Dropped**
- `cpt-store.ts`, tracker/node wiring, and the flag. Live-traffic savings were never A/B-measured, and repo policy treats opt-in flags with no caller as dead code. If a shadow-mode A/B shows the learned tables beat the recalibrated constant, the store wiring is worth revisiting on top of `fitCpt`.

Thanks — the decomposition model and the fail-closed guard design were the hard parts, and they're in as-is.

PR #129, opened 2026-07-20
fix(responses-bridge): encode assistant text as output_text, not input_text

dex0shubham:
### Problem
When a Claude Code `/v1/messages` request is bridged to a GPT model on the Responses path (`bridgedGptMessages`, `proxy.ts`), prior **assistant** text turns were encoded as content parts of type `input_text`. The OpenAI Responses API only accepts `output_text` (or `refusal`) for `role:assistant` content — `input_text` there is a hard **400**, so any multi-turn conversation to a Responses-routed GPT model failed after the first assistant reply.

The bug is confirmed by the bridge's own asymmetry: the read side (`openAIResponseToAnthropicMessage`) already maps assistant text ↔ `output_text`; only the write side (`inputParts`) was wrong. The branch had no test coverage.

### Fix
Thread the message `role` into `inputParts` and select the text part type: `output_text` for assistant, `input_text` for user/system (unchanged). Added a regression test in `responses-bridge-roles.test.ts`.

Full suite passes (804/804); `typecheck` clean.

### DIFF
--- a/src/core/messages-responses-bridge.ts
+++ b/src/core/messages-responses-bridge.ts
@@ -36,14 +36,18 @@ function imageUrl(source: unknown): string | undefined {
   return undefined;
 }
 
-function inputParts(content: unknown, location = 'message'): JsonObject[] {
-  if (typeof content === 'string') return [{ type: 'input_text', text: content }];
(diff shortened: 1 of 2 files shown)


teamchong (approved):
Thanks

PR #131, opened 2026-07-21
fix(responses-bridge): treat empty tool-call arguments as {} instead of crashing

dex0shubham:
### Problem
On the non-streaming Responses to Messages conversion (`openAIResponseToAnthropicMessage`), a `function_call` with `arguments: ""` (what some models emit for a no-argument tool call) reaches `JSON.parse("")`, which throws. The error propagates out and fails the whole turn.

This is inconsistent with two sibling paths that handle it fine:
- The Chat bridge guards it explicitly (`value.trim() === '' -> return {}`, `messages-chat-bridge.ts`).
- The Responses **streaming** path already yields `input: {}` for empty args.

### Fix
Mirror the Chat bridge's guard in the Responses `parseArguments`: empty/blank string maps to `{}`. Added a regression test (the non-streaming converter had no direct coverage).

Full suite passes (807/807); `typecheck` clean.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/core/messages-responses-bridge.ts
+++ b/src/core/messages-responses-bridge.ts
@@ -200,7 +200,11 @@ function anthropicUsage(raw: unknown): JsonObject {
 }
 
 function parseArguments(value: unknown): unknown {
-  if (typeof value !== 'string') return {};
+  // Empty string = a no-argument tool call; treat as {} rather than throwing.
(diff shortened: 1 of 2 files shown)


teamchong (approved):
Thanks

PR #144, opened 2026-07-23
Cache-churn fix (sliding prefix), model-scope persistence, dashboard i18n + observability, Windows launcher — also carries a rebased copy of #147

SiNaPsEr0x:
Hi! This PR upstreams the fixes I've been running on my fork, rebased on top of your current `main` (`8a9175d`). 47 test files / 862 tests pass and `tsc --noEmit` is clean.

> ### ⚠️ Please read this first — the scope of this PR has drifted
>
> This PR is opened from my fork's `main` branch rather than from a dedicated topic branch, so everything I've pushed to `main` since opening it has landed here automatically; the diff is now 25 commits / 83 files.
>
> The reason it grew this way is timing. New models ship faster than this PR could get reviewed, and each one needed profile and reader work stacked on top of the previous one. Rather than let the fork stall waiting, I integrated everything into my `main` in one go — it has been running there since, and it works. If PRs here could get a look sooner, I'd be able to keep them small and single-purpose; as things stand, keeping them separate would have meant keeping them all unmerged.
>
> Most importantly, **five of the commits in this PR are authored by you**, not me — `f95e90c`, `a51fa7e`, `9d50648`, `8d21fdb`, `df90662`. They are a rebased copy of your own **#147** (claude-opus-5 as default reader profile), which I merged into my fork while working on top of it. They carry **different SHAs** than the ones on `feat/native-14px-reader-profile`, so Git will not recognise them as already applied — merging this PR as-is would pull in #147's content under new hashes.
>
> An earlier version of this description said "the diff is only my patches". That was true when I wrote it and is not true anymore; I'm correcting it here rather than leaving it stale.
>
> **If you'd prefer to review only my own changes, say so and I'll re-open this from a clean branch with those five commits dropped.** That's what I'd recommend — it would take the diff down to my patches alone and remove the overlap with #147 entirely. I didn't want to force-push a rewritten history under an open review without asking first.

## 1. Fix cache churn / "sliding cache" — the big one (`a058439`)

This fixes the silent cache-loss problem. Two root causes:

- **`total_tokens` in the static slab.** The `<total_tokens>N tokens left</total_tokens>` block decrements every turn. Since it sat inside the cacheable static prefix, it broke the prefix right after itself on every request: verified via A/B in `events.jsonl` — `cache_read` stayed frozen at the pure-slab size while `cache_create` re-billed the whole tail at 1.25x instead of 0.1x reads. Fixed by adding `total_tokens` to `DYNAMIC_BLOCK_TAGS` so it's carved out of the prefix.
- **The relocated env wrapper rode a *moving* slot.** The volatile env/context text was appended to the *last* user message, but that slot roams forward every turn (pxpipe is stateless and the client never sees the wrapper, so at turn N+1 the message that carried it at turn N is replayed *without* it). Byte-identical content at a moving position still diverges the replayed prefix at the old position — so every cache entry downstream of the slab/history anchor missed, and the entire live tail was re-created at 1.25x each turn (live data 2026-07-21: read frozen at 25669 for 8 turns while create climbed 8797→21312). Fixed by pinning the wrapper to a **position-stable slot** (end of the *first* user message). The wrapper bytes are turn-invariant, so the tail now accrues into cache; a genuine env change (version bump) costs one full-prefix rewrite, which is the accepted trade-off.

Also adds prefix-change tracking in `src/core/tracker.ts` so future regressions of this kind are observable instead of silent, plus e2e coverage in `tests/cache-stability-e2e.test.ts` and `tests/tracker.test.ts`.

## 2. Persist model-scope selection + validate model base ids (`5e488c2`)

Makes the proxy work reliably across **all model families** (Claude / GPT / Gemini / Grok bases), not just the default:

- New `src/config-file.ts` (`resolveConfigFilePath`, `persistModelsToConfigFile`) shared by `node.ts` startup and the dashboard.
- A model chip toggled in the dashboard now survives a process restart instead of resetting to the env/built-in default.
- `isValidModelBaseId()` shape-check before persisting, so an unauthenticated local POST can't write arbitrary junk into `config.json`.
- Default config path corrected to `~/.pxpipe/config.json` (matching the dir already used for `events.jsonl`); `--help` updated.
- Tests point `PXPIPE_CONFIG` at a per-test tmpfile so they never touch a developer's real config.

## 3. Dashboard i18n: EN/IT language selector + polish (`10faa51`, `7fec329`)

- Live dashboard gets an EN/IT language selector (Accept-Language default, persisted choice).
- Copy polish ("Ispezione immagine", "Hit di cache") and removal of the obsolete Fable-5-only compression warning, which no longer applies now that model scope is persisted per-family.

## 4. pnpm overrides → `pnpm-workspace.yaml` (`90b6308`)

Moves overrides out of `package.json` into `pnpm-workspace.yaml`, the supported location on current pnpm — fixes install warnings and keeps overrides applied.

## 5. Restart script: resolve Git Bash explicitly (`6b2e439`, `3d66a9a`)

On Windows, `restart.sh` no longer relies on PATH order (which could pick WSL bash or fail); it resolves Git Bash explicitly. Plus `.gitattributes` for line-ending hygiene.

## 6. Windows launcher + claude-mem compatibility wrapper (`1c7776c`, `68ff026`, `bccdcba`)

- New `claude-pxpipe.ps1` launcher: starts pxpipe if not already running, sets `ANTHROPIC_BASE_URL` for the session only (no global env pollution), and launches Claude Code through the proxy.
- `claude-noproxy` wrapper (`claude-noproxy.cs`, compiled on the fly with the .NET Framework `csc` compiler bundled with Windows — no extra toolchain needed): lets the claude-mem plugin bypass the proxy and talk to the API directly, avoiding the claude-mem/proxy conflict. The launcher detects a claude-mem install and points its `CLAUDE_CODE_PATH` at the wrapper automatically; on other platforms the check is skipped.
- README documentation for the launcher modes (`-Local`, `-DebugCapture`, the `npx` caveat) and the claude-mem noproxy wrapper.

## 7. Dashboard: rate-free savings pair + static-tag churn observation (`98eefab`, `d034aa0`)

- Reports the savings pair without baking in per-token rates, so the figure doesn't silently rot when pricing changes.
- Surfaces static-tag churn as an observation, which is what made the two bugs in item 1 visible in the first place.
- Reframes the model-scope warning around *what gets read*, rather than cost — the previous copy implied a billing claim it couldn't support.

## 8. Skip the redundant cacheable-prefix probe (`92040cd`)

When the cacheable-prefix probe body is byte-identical to the baseline body, the second upstream round-trip is pure overhead. Now it reuses the first result instead. Covered by `tests/proxy-usage.test.ts`.

## 9. Normalize CRLF so `droppedChars` stays a real signal (`e9d3d82`)

On a CRLF checkout, `minifyForRender` counted every stripped `\r` as a dropped character, so the render-quality signal read as permanent low-level data loss on Windows while nothing was actually lost. Line endings are normalized before the count, leaving `droppedChars` meaningful as a genuine-loss indicator.

---

Happy to split this into separate PRs if you prefer smaller reviews — and see the note at the top about dropping your five #147 commits from this branch. The cache fix (item 1) is the one I'd most like to see land, since every long session silently loses caching without it. Thanks a lot for the project and for taking a look! 🙏


teamchong:
Thanks, but I can't reproduce this on my side
my events.jsonl (54,809 events) has zero occurrences of `total_tokens`, and my cache read/create ratio sits at 8-17x across the last week. A broken prefix should put that below 1.

SiNaPsEr0x:
> Thanks, but I can't reproduce this on my side my events.jsonl (54,809 events) has zero occurrences of `total_tokens`, and my cache read/create ratio sits at 8-17x across the last week. A broken prefix should put that below 1.

Fair point on the aggregate ratio — but the zero-total_tokens result likely says more about your Claude Code setup than about the bug's existence. Vanilla Claude Code with no plugins at all is increasingly the minority case now; most active setups run at least one plugin injecting a status/context indicator into the prefix, which is exactly where total_tokens/severity/category came from in the reports so far. If your traffic has none of those, you're not hitting the precondition — that doesn't rule out the mechanism, it just means your setup doesn't trigger it.

To be precise about scope: the fix doesn't blanket-immunize against every plugin that will ever exist. It moves the tags identified so far into DYNAMIC_BLOCK_TAGS, and adds prefix-change tracking (tracker.ts) so a new plugin-injected tag that breaks the static prefix shows up as an observable regression instead of a silent one — still needs to be added to the list when it's found, but it won't be silent anymore.

SiNaPsEr0x:
I send mirate update now  in other PR

PR #147, opened 2026-07-25
feat: add claude-opus-5 as a default reader; keep prompts and rules as text

teamchong:
`claude-opus-5` joins the default `PXPIPE_MODELS` set, and the collapse path
stops rasterizing the two things a model must not misread: the user's own
prompts and the standing rules in `CLAUDE.md`.

## Opus 5 as a default reader

It reads rendered pages about as well as Fable 5 except at character-level
transcription, and holds ~4.7x more text per context window than plain text.

| test | Fable 5 | Opus 5 |
|---|---:|---:|
| arithmetic on image-only numbers | 100/100 | 100/100 |
| gist recall | 98/98 | 94/98 |
| state tracking | 18/18 | 17/18 |
| hallucinated facts (lower better) | 0/16 | 0/16 |
| exact transcription, dense string | 13/15 | 2/15 |

The transcription row is the one that matters: Opus 5 cannot separate glyphs
reliably at our render density. Keep Fable 5 or Gemini 3.6 Flash for verbatim
work. Effort defaults to `medium`. The failures are perceptual, so more
thinking does not recover the glyphs.

## Why default it anyway

Opus 5 is not as good as Fable 5 at reading, and I am shipping it as a default
regardless. The reasons, plainly:

- It passes `demo/effective-context/b.sh`.
- I dogfooded it for a day and it completed real tasks.
- Sessions that would run past 1M tokens and die as text now complete as
  images. I can work far longer before reaching for `/compact`.

Someone will ask how one day of use tells me whether this is better or worse
than running without pxpipe. It does not. I have not measured that, and I am
not claiming it. Every tradeoff already written down in the README still
applies, including the transcription gap above.

If it turns out in practice that Opus 5 support is much worse, I am happy to
revert this.

## Tasks completed past 1M context
image

## Prompts and rules stay text

Assistant prose and tool output are reconstructable from the work itself. A
user instruction and a `CLAUDE.md` rule are not: they exist only in the
transcript, and they are exactly what an OCR slip corrupts silently.

Collapse now renders a user message *minus* its typed text blocks and re-emits
those blocks as native text beside the image. `tool_result`s in the same
message still render, so they stay adjacent to the imaged `tool_use` they
answer. Past 2000 chars a "prompt" is a pasted document rather than an
instruction, so it gets its own image instead of bloating the text block, and it
still never joins the history transcript.

```
Before:  <user t="12">  ...prompt...  </user>   (pixels)
After:   [User turns from this session, verbatim ...]
         <user t="12">...prompt...</user>       (text)
```

`CLAUDE.md` gets the same treatment via `stripMarkdownEnvSection` +
`extractClaudeMdSlab`: the rules ride out as text on the first user message,
never re-imaged on later turns.

Benchmarks and receipts: `FINDINGS.md`, `eval/gsm8k/`, `eval/gist-recall/`,
`eval/verbatim-15/`.


Artemonim:
close #149

PR #152, opened 2026-07-26
Make the repo work on Windows without WSL (and fix a <total_tokens> cache invalidation bug)

SiNaPsEr0x:
## Summary

Makes the repo usable on Windows without WSL, and fixes one platform-independent
cache bug found along the way.

On a clean Windows clone, `pnpm run restart` failed immediately:

```
scripts/restart.sh: riga 27: $'\r': comando non trovato
scripts/restart.sh: riga 115: exec: node: non trovato
```

Two separate causes: the repo had no `.gitattributes`, so shell scripts were
checked out with CRLF; and `restart.sh` relied on `pgrep`/`kill`/`exec` plus a
POSIX-only PATH lookup.

## Changes

**`.gitattributes` (new)** — normalizes line endings so checkouts are consistent
across platforms, and forces LF on files carrying a shebang so they stay
executable when cloned from Windows. Binary assets (`.png`, `.ico`, `.woff2`, …)
are marked explicitly to avoid corruption.

**`scripts/restart.sh` → `scripts/restart.mjs`** — reimplemented in Node, no
shell dependency. Process discovery and termination now go through
platform-appropriate paths (`ps`/`kill` on POSIX, `Get-CimInstance` +
`Get-NetTCPConnection` on Windows), with the same graceful-then-forceful
shutdown sequence and the same port-holder diagnostics as before. Behaviour on
Linux/macOS is unchanged.

**`tests/restart.test.sh` → `tests/restart.test.ts`** — the restart suite now
runs under vitest on every platform instead of requiring bash. This also drops
the separate `test:restart` script, since `pnpm test` covers it.

**`tests/node-security.test.ts`** — POSIX permission-bit assertions are skipped
on Windows, where the filesystem has no equivalent concept. The assertions still
run everywhere else.

**`src/core/transform.ts`** — `<total_tokens>` added to `DYNAMIC_BLOCK_TAGS`.

## The cache bug (not Windows-specific)

`<total_tokens>` was being treated as static content. It changes on every single
turn, so it was landing in the static slab and rewriting the cached image each
turn — dropping the cache hit rate to zero for any conversation long enough to
reach the image path. It also surfaced as recurring `unknownStaticTags` noise in
the logs, which is what led me to it.

This one is worth reviewing independently of the Windows work: it affects every
platform.

## Testing

Full suite green on Windows: **923 tests across 51 files**. The restart suite is
included in `pnpm test` now, so it runs in CI on all platforms rather than only
where bash is available.

I don't have a macOS or Linux box to verify on, so a second pair of eyes on the
POSIX branch of `restart.mjs` would be welcome — the logic is a direct port, but
it is the part I could not exercise myself.


teamchong:
Landed in ef3c97c0 with you as commit author — thanks for pushing on this!

Your diagnosis was exactly right: the bash script was unrunnable on Windows (no bash/pgrep/lsof, plus CRLF corruption on checkout). I reworked the implementation as a Node script so both platforms share one code path:

- process discovery: `ps` on POSIX, `Get-CimInstance Win32_Process` on Windows
- graceful stop: SIGTERM / `taskkill` (no /F) with a 5s grace period, then force
- port probe with holder diagnostics (`lsof` / `Get-NetTCPConnection`)
- helpers are exported with injectable deps, so tests/restart.test.ts covers the sequencing without touching the real process table
- .gitattributes pins LF for scripts so Windows clones stop breaking bash and shebang lines

`pnpm restart` now runs `node scripts/restart.mjs` everywhere. Closing since the change landed via main rather than this branch.

PR #162, opened 2026-07-30
feat(render): uppercase A-Z inked blue for case parity (default on)

davidguo163:
## What

Uppercase A–Z now carries a fixed blue ink (`UPPER_INK = [0,70,190]`) on every imaged channel — case parity for shape-identical pairs. Lowercase, digits, and punctuation keep black ink; the `l/i/I/o/O` reading rules and the red `↵` marker are unchanged.

At 5×8, `w/W`, `c/C`, `s/S`, `x/X`, `z/Z` differ only by scale, and after the vision encoder resamples, scale is gone — color is the only channel left. A real `w→W` misread shipped in long-context; 200/200 single-image probes never reproduced it, and the owner approved implementing the defense without a red repro.

## How

- `render.ts`: new `RenderStyle.upperBlue` (default **on**; `false` opts out). Rides the existing colorMask/palette machinery — no slot-string, wrap, or shrink changes.
  - Pages with **no** A-Z stay byte-identical grayscale (colorMask not allocated), so the cache/byte-freeze story is intact for pages the feature doesn't touch.
  - Composes with `colorByRole` (role tags keep their hue; uppercase body glyphs go blue — deliberately NOT the assistant-role blue so both signals stay tellable apart) and with `markerRed` (red `↵` keeps priority on the RGB path). `colorCycle` wins where explicitly set.
- Banners (`transform.ts` slab header, `openai.ts` CHAT/RESPONSES headers): document the new semantics — "UPPERCASE letters are inked blue, lowercase stays black — when a letter's case is ambiguous (w/W, c/C, s/S), trust the color."
- Profiles: `GptRenderStyle.upperBlue` wired through `BASE_STYLE`, the Gemini profile, and `PXPIPE_MODELS` env-override parsing, so it can be disabled per model.

## No-regression evidence

- `pnpm run typecheck` clean; `pnpm test` **934/934** green (52 files), including 14 new targeted tests (`tests/upper-blue.test.ts` + `tests/render.test.ts`): uppercase-blue/lowercase-black per-cell ink assertions, w/W-style pair distinguishability, l/i/o/I/O/digit ink semantics preserved, byte-identity for no-uppercase pages and for `upperBlue:false`, pixel diff confined to uppercase cells (no layout drift), `colorByRole` + `markerRed` composition.
- Out-of-band golden check: rendered a 22-page corpus (render.ts source, mixed-case, lowercase-only, CJK-mix) on base `d41f056` vs this branch, decoded both PNG sets, and pixel-diffed 22.8M pixels: **86,553 differing px, 0 non-blue** — every changed pixel is blue-dominant uppercase ink; dimensions, wrap, page count all byte-stable. Lowercase-only pages: **0 differing pixels**. `upperBlue:false` reproduces base **bit-for-bit** on all 22 pages.
- Size (recorded, not gated): grayscale→RGB flip costs bytes on pages that contain uppercase, e.g. dense code page 13.5 KB → 24.1 KB (+78%), CJK-mix page 4.5 KB → 52.7 KB (worst case: RGB + colored ink defeats the gray palette). Billed vision tokens are unchanged — patch-grid billing is dimension-only, and dimensions are untouched.

Scope guard: no changes to fold/wrap anchors, slot/lockstep mechanism, or `shrinkColsToContent`/degradation paths.

### DIFF
--- a/src/core/gemini-model-profiles.ts
+++ b/src/core/gemini-model-profiles.ts
@@ -41,6 +41,7 @@ export const GEMINI_3_6_FLASH_PROFILE: GptModelProfile = {
     markerScale: 1,
     markerRed: false,
     inkDilate: 0,
+    upperBlue: true,
   },
 };
 
--- a/src/core/gpt-model-profiles.ts
+++ b/src/core/gpt-model-profiles.ts
@@ -67,6 +67,9 @@ export interface GptRenderStyle {
   markerRed: boolean;
   /** Pre-invert ink dilate radius (px). 0 = off. Thickens glyphs at fixed cell pitch. */
   inkDilate: number;
+  /** Uppercase A-Z inked blue for case parity (w/W, c/C are shape-identical at 5×8).
+   *  Default true on every imaged channel; set false to opt out per model. */
+  upperBlue: boolean;
 }
 
 export interface GptHistoryProfile {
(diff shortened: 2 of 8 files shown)


davidguo163:
Withdrawn by owner decision — no external contributions to pxpipe.

PR #166, opened 2026-08-01
Add Novita AI as a provider option

jax-novita:
## Summary

- Adds Novita AI as a provider option.

## Validation

Compared against the baseline on `d41f0566f10c`: compared_to_baseline. Pre-existing failures are left as-is.

## Reviewer Notes

- **This is a documentation-only change; no code path is added or modified.** Novita's inference API is OpenAI-compatible and routes through the `OPENAI_UPSTREAM` / `OPENAI_MODELS` mechanism this project already ships and already documents for OpenAI itself (and, via Cloudflare, for Kimi K3). We did not add a new provider abstraction.
- **The routing was verified reachable, not functionally exercised.** We do not hold a Novita API key. `curl` against `https://api.novita.ai/openai/v1/responses` and `.../v1/chat/completions` both return `403 INVALID_API_KEY` rather than `404`, which confirms the endpoints exist at the documented shape, but we did not send an authenticated request and confirm a real completion comes back through pxpipe end to end.
- **Model ID in the example (`deepseek/deepseek-v3.2`) was pulled from Novita's live `/v3/openai/models` listing** at the time of writing, not guessed; it may not be the model you'd choose to feature.
- **Token-compression eligibility (`PXPIPE_GPT_PROFILES` / render profile) is intentionally out of scope.** Making a Novita model participate in this project's image-compression feature requires a measured vision-token cost profile (as every other model in `src/core/gpt-model-profiles.ts` has), and we had no way to measure that without an API key. Listing a placeholder number would corrupt the profitability gate silently. This PR only adds the routing recipe; a model added this way runs uncompressed until someone measures and contributes a profile, same as any other newly-routed OpenAI-compatible model on day one.
- Full test suite (`pnpm test`, 51 files / 920 tests) passes identically before and after this change, which is expected for a docs-only diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ behavioral changes, patch = fixes).
 
 ## Unreleased
 
+### Added
+- Documented routing to Novita's OpenAI-compatible endpoint via the existing
+  `OPENAI_UPSTREAM` / `OPENAI_MODELS` mechanism (no new code path).
+
 ### Changed
 - Opt-in `gpt-5.6-sol` now uses native 14px JetBrains Mono at 84 columns. Its paid
   pilot preserved gist and guard checks, read 7/8 exact values, and produced no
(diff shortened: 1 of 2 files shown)


teamchong:
Merged as a community-contributed example. To be clear: I don't use Novita and this isn't a partnership or endorsement — it's kept because it's a correct, honest illustration of the generic `OPENAI_UPSTREAM` mechanism, and the section is now labeled as such in the doc. Future vendor PRs should follow the same format: docs-only, clearly labeled, no token-profile changes. Thanks for the straightforward disclosure about what was and wasn't tested.

Issue #38, opened 2026-07-04
Render in-image exact strings as checksummed wordlist encodings instead of raw hex


aeoess:
Your legibility audit already contains the shape of a fix. The miss taxonomy is 100% predicted by the glyph-confusability matrix, hex blind-reads at 3/8 because `0~O` `3~8` `5~S` `8~B` sit within a couple of pixels of each other, and the factsheet covers SHAs that can ride as text. The residual problem is byte-exact values that have to stay inside the rendered page (inline hashes and IDs in collapsed history), where a misread is a silent confabulation.

Proposal: at render time, re-encode those spans as words from a fixed lexicon instead of raw hex. A 2048-word list carries 11 bits per word, so a 48-bit id is 5 words plus a checksum word. Two properties fall out. First, single-glyph substitutions almost never land on another valid word, so the same vision noise that silently flips `3` to `8` produces an out-of-lexicon token the model can flag. Second, the checksum word makes the failure mode "detected mismatch" instead of "confident wrong read", which is the difference your Opus 0/15 result needs.

This is a solved encoding problem from another noisy channel: the PGP word list was designed for reading key fingerprints aloud over a phone line, with even/odd word alternation as position parity, and BIP39 adds the checksum pattern. Vision reading of a rendered page is the same channel class as a phone line, just with a glyph confusability matrix instead of a phonetic one.

Cost is roughly 3 to 4x the glyph footprint per encoded span, which is noise against a 92k-char page. The render side needs a regex pass for hex-shaped spans plus a lookup table; the read side needs one line in the OCR banner telling the model the encoding exists and to report checksum failures instead of guessing.

I can draft the encoder plus an eval in your house style: n random 64-bit handles rendered dense, hex arm vs wordlist arm, per model, misreads split into silent vs detected. The silent-to-detected conversion rate is the number that matters.

teamchong:
yes, thank you, welcome to the visual context compression experiment

I suggest to use BIP39 or the PGP word list, not a custom one, the model has to recognize the words without a lookup table in context

Artemonim:
If I understand the method correctly, it seems to me that it could add unnecessary cognitive cost to the prompt. The current implementation with footnotes seems more optimal.

aeoess:
Numbers, finally.

I ran the comparison at the audit's shipped layout: 1568x728, Spleen 5x8, 40 random 64-bit values per arm, exact-match scoring. The harness and manifests were pinned before each run.

[...]

PR #179, opened 2026-08-04
wordlist: BIP39 checksummed word encoding for high-entropy spans (refs #38)

davidguo163:
Refs #38 — this is the encoder half of that proposal, in-repo with tests.

## Why

Hex is the worst glyph class under vision readback: a `3`→`8` flip lands on another valid value, so the failure mode is a *confident wrong read*. Encoding the same bytes as words from the standard BIP39 lexicon (per the guidance in #38: use a lexicon models already know, not a custom one) gives:

1. single-glyph substitutions almost never form another valid word → out-of-lexicon token the reader can flag;
2. the SHA-256 checksum turns residual valid-word substitutions into **detected mismatch** instead of a silent flip.

Measured on a readback bench (64-bit handles rendered dense, hex arm vs wordlist arm, transcribe + decode, misreads split silent vs detected):

| model | hex arm | wordlist arm |
|---|---|---|
| claude-fable-5 | 1 silent error, **16 missing** values | 0 silent, 0 missing (5 misreads all caught by checksum) |
| gemini-3.6-flash | 3 silent errors | 0 silent (3 misreads all caught by checksum) |

i.e. the silent→detected conversion rate #38 asks for was 100% in both arms; words also transcribe more completely than hex for the weaker reader (16 dropped values → 0).

## What

- `src/core/wordlist.ts` — `wordlistEncode(bytes)` / `wordlistDecode(words)`; generalized BIP39: ENT-bit payload (ENT % 32 = 0) + first ENT/32 bits of SHA-256 as checksum, 11 bits/word. A 64-bit handle → 6 words. Decode throws typed `WordlistDecodeError` (`unknown-word` / `bad-word-count` / `checksum-mismatch`) instead of guessing. SHA-256 via `crypto.subtle` (Node + Workers), zero deps.
- `src/core/bip39-words.ts` — the standard 2048-word BIP39 English list.
- `tests/wordlist.test.ts` — reference BIP39 128-bit vectors (at 16/20/24/28/32 bytes the scheme is exactly standard BIP39), round-trips for all supported sizes, glyph-corruption → `unknown-word`, and a deterministic valid-word-substitution sweep asserting detection ≥ 70% at 2 checksum bits / ≥ 90% at 4 (and never a silent decode back to the original).

`npx vitest run tests/wordlist.test.ts` 8/8, `tsc --noEmit` clean.

Deliberately **not** wired into the render path yet — per #38 the remaining halves are the render-side regex pass for hex-shaped spans (factsheet already classifies `hex`/`uuid` shapes) and the OCR-banner line telling the reader to report checksum failures. Happy to follow up with either if you want this direction; also fine if you'd rather take just the scheme and reshape the API.


davidguo163:
Withdrawing this PR per our side's decision — keeping the implementation in our fork only. Data/analysis in the PR description remains available if useful for #38.

PR #217, opened 2026-08-09
fix(transform): degrade legible history to dense before abandoning the collapse (#216)

### LINKED ISSUE #216: 0.13.0: legible profile default for claude-opus-5 breaks long sessions - collapse bails, raw multi-megabyte history sent as text, upstream 400s
# 0.13.0: legible profile default for claude-opus-5 breaks long sessions - collapse bails, raw multi-megabyte history sent as text, upstream 400s

## Summary

Since 0.13.0 (#170), `claude-opus-5` defaults to the *legible* history geometry (172 cols, `jetbrains-mono-14`) instead of the dense one (`spleen-5x8`, 312 cols). Only collapsed history changes; the static slab and tool-result pages stay dense. On long sessions this makes the rendered history ~7x larger per character (production median 6.7 to 48.3 `image_bytes`/`collapsed_chars`, see below). Once the render exceeds the image budget, the atomic-admission check (`transform.ts:1991-1998`, `2579-2586`) abandons the collapse **entirely** and forwards the raw history as plain text - observed `outgoing_text_chars` 2.43-2.56M, which the proxy's own `count_tokens` probe measured at 1.65-1.73M tokens (`baseline_tokens`, `baseline_probe_status: "ok"`).

Forwards of that size are rejected by the upstream API with 400: of the 8 bail events, the 6 measuring 1.65-1.73M tokens returned 400, while the two that stayed at 825k/832k chars (619k/621k tokens) completed normally (`end_turn` / `tool_use`). No 4xx response body was captured, so the threshold is bounded by observation rather than read off an error message.

The turn was re-attempted repeatedly - 4 distinct request bodies (`req_body_sha8`), two of them sent twice, gaps 9.7-186 s, 7.2 minutes end to end. From the second bail onward the session was already on the cold-cache repack path (`history_pack_fill: true`), which produced 400s of its own (see below).

Note on budgets: the admission check runs against `imageByteHeadroom` = 18 MiB minus the caller's native image bytes and the image bytes already admitted for this request (slab, tool-result pages) (`transform.ts:1847`, limit at `transform.ts:158`), so the logged `image_bytes` of the 400 events (1.72-1.77 MiB) is consumption *without* the abandoned history render.

## Environment

- pxpipe-proxy 0.13.0 (regression; window is clean-cut, see Evidence); since downgraded to 0.11.1
- Node v26.4.0, Linux, systemd service
- `PXPIPE_MODELS=claude-fable-5,claude-opus-5`

## Evidence (from `~/.pxpipe/events.jsonl`)

- Regression window is clean-cut: **0** `history_reason: "image_bytes"` events in 97.5k logged requests (2026-07-04 through the restart into 0.13.0 at 14:50:50Z); first bail 32 s after that restart. After downgrading to 0.11.1 at 15:22:35Z: 508 requests up to 16:04Z, 234 of them `claude-opus-5` with `orig_chars` > 200k, 0 x 4xx (three transient upstream 502s, all on normal-sized requests).
- All primary-class events carry `history_reason: "image_bytes"` - 8 bail events total: 6 x 400, 2 x 200.
- Median render cost `image_bytes`/`collapsed_chars` for `claude-opus-5`: **6.7** (n=44,411 requests before the restart) to **48.3** (n=85 requests on 0.13.0), measured at the exact restart into 0.13.0.
- Failing requests: `outgoing_text_chars` 2.43-2.56M sent uncollapsed (`orig_chars` of the static slab stays ~204-208k); the `count_tokens` probe puts them at 1.65-1.73M tokens.
- Secondary failure class: 8 x 400 with `history_reason: "collapsed"` and normal-looking `outgoing_text_chars` (69.7-139.7k), 14:55:34Z-15:14:58Z, 6 of them carrying `history_pack_fill: true` - all after the first bail and interleaved with the tail of the primary class. No 4xx body captured, so cause unconfirmed.

## Expected behavior

If the legible geometry blows the image budget, degrade to the dense profile (or split/chunk the render) instead of abandoning the collapse. Sending 1.65-1.73M tokens of raw text as the fallback cannot succeed - bailing must not produce a request larger than the render it refused to send.

## Repro

1. Session where the legible render of the collapsed history exceeds the 18 MiB headroom with `claude-opus-5` on 0.13.0 defaults (in our case the abandoned raw forward was 2.4-2.6M chars; the static slab ~205k chars).
2. Watch events.jsonl: `history_reason: "image_bytes"`, then plain-text forward, upstream 400.

## Workarounds

- Downgrade to a pre-legible-default version (we run 0.11.1) - verified in production, numbers above.
- A `PXPIPE_GPT_PROFILES` entry for the model id should equally avoid the bail - the env map is consulted before the built-in Claude table (`gpt-model-profiles.ts`, `resolveGptProfile`) and accepts both fields: `PXPIPE_GPT_PROFILES='{"claude-opus-5":{"historyStripCols":312,"historyStyle":{"font":"spleen-5x8"}}}'`. Set both together - omitted fields inherit from the legible built-in. Not verified in production.

Related: #210 (frozen history re-rendered every turn - the legible default multiplies that cost as well).

ousamabenyounes:
## Summary

Since v0.13.0 (#170), misread-prone Claude models (e.g. `claude-opus-5`) default to the *legible* collapsed-history geometry (`jetbrains-mono-14`, 172 cols), which renders several times heavier per char than the dense geometry (`spleen-5x8`, 312 cols). On long sessions the legible history render exceeds the 18 MiB image-byte headroom. The atomic-admission check (`transform.ts`) then **abandoned the collapse entirely** and forwarded the raw history as plain text — a request *larger* than the render it refused (production `outgoing_text_chars` 2.43–2.56M ≈ 1.65–1.73M tokens), which the upstream API rejects with 400.

Per the issue's expected behavior: *"If the legible geometry blows the image budget, degrade to the dense profile instead of abandoning the collapse. Bailing must not produce a request larger than the render it refused to send."*

## What changed

- New shared helper `collapseHistoryWithinByteBudget`: collapse at the model geometry, and when that render overflows the byte budget **on a model that renders legible history**, retry once at the dense geometry. Keep the dense collapse only if it fits; a history too big for dense too still keeps its text, so the atomic-admission guarantee is preserved.
- Both previously-duplicated admission sites (`runHistoryCollapseAndFinalize` and `transformRequest`) now route through the helper — the bug existed in both.
- Surface the fallback as `history_degraded_dense` in `events.jsonl` (tracker) for the same observability the issue was diagnosed with.
- CHANGELOG entry under 0.13.0 `### Fixed`.

Dense-native profiles (Fable, GPT) skip the retry — they have no legible geometry to degrade from, so re-rendering would just reproduce the overflow.

| | |
| --- | --- |
| Issues | Fix #216 |
| Base | `main` |

## Test verification (RED → GREEN)

Real runs on an unmodified upstream base vs the patched branch (measurement-driven fixture in `tests/image-byte-budget.test.ts`). The fixture measures the dense and legible render sizes with a non-binding budget, then transforms `claude-opus-5` with a budget **between** the two: legible overflows, dense fits.

**RED** — new tests on unmodified `transform.ts` (bug present): the legible collapse overflows and is abandoned to a raw-text forward (`historyReason: 'image_bytes'`).

```
 × a legible history that overflows degrades to dense before abandoning the collapse (#216)
     > re-renders the collapse at dense instead of forwarding raw text (imaged-slab path)
 × a legible history that overflows degrades to dense before abandoning the collapse (#216)
     > re-renders the collapse at dense instead of forwarding raw text (text-slab path)
 Tests  4 failed | 9 passed
```

Direct probe of the bug at a 5 MB budget on unmodified main:
```
opus-5 (legible), maxImageBytes=5,000,000
  historyReason = image_bytes     (collapse abandoned)
  outgoingTextChars = 929,804     <-- raw history forwarded as TEXT
```

**GREEN** — same tests with the fix: the collapse is re-rendered at dense and admitted.
```
 ✓ ... re-renders the collapse at dense instead of forwarding raw text (imaged-slab path)
 ✓ ... re-renders the collapse at dense instead of forwarding raw text (text-slab path)
 ✓ ... leaves Fable (already dense) on the original atomic-admission bail
 ✓ ... still bails when even the dense render overflows the budget
 Tests  2 failed | 11 passed
```
(the 2 remaining failures pre-exist on `main` unchanged — see below)

Same 5 MB probe with the fix:
```
opus-5 (legible), maxImageBytes=5,000,000
  historyReason      = collapsed
  historyDegradedDense = true
  imageBytes         = 2,860,986   (dense render, within budget)
  outgoingTextChars  = 176,410     (was 929,804 raw text)
```

The bail invariant now holds: what we forward is never larger than the render we refused.

## Full local CI (`pnpm install --frozen-lockfile` · audit · typecheck · test · build)

- **Baseline (pristine `main`):** `2 failed | 1151 passed` — typecheck/build green.
- **This branch:** `2 failed | 1155 passed` — typecheck/audit/build green.

Iso-better: same 2 pre-existing failures, +4 new passing tests, no new failures. The 2 pre-existing failures (`keeps a tool_result as text when its pages do not fit`, `warns before the next turn walks into the wall` in `tests/image-byte-budget.test.ts`) fail identically on pristine `main` with this change reverted — they assert hardcoded byte thresholds sensitive to the local canvas renderer and are unrelated to this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,12 @@ behavioral changes, patch = fixes).
   models (#159).
 - `build.mjs` resolves `tsc` via `typescript/package.json` instead of the
   removed `./bin/tsc` export, fixing builds on newer TypeScript.
+- When the legible history geometry (#170) overflowed the image-byte budget on a
+  long session, the collapse was abandoned and the raw history forwarded as
+  text — a request larger than the render it refused, which the upstream 400s.
+  The collapse now degrades to the dense geometry to fit instead of bailing;
+  only a history too big for dense too keeps its text. New `history_degraded_dense`
+  event marks the fallback (#216).
 
 ## 0.12.1 — 2026-08-08
 
--- a/src/core/tracker.ts
+++ b/src/core/tracker.ts
@@ -93,6 +93,10 @@ export interface TrackEvent {
    *  was therefore allowed to repack for density. Pair with cache_read_tokens:
    *  a repack that lands on a live cache would show as a cache_create spike. */
   history_pack_fill?: boolean;
+  /** Set when the model's legible history render overflowed the image-byte
+   *  budget and the collapse was re-rendered at the dense geometry to fit,
+   *  rather than abandoned to a raw-text forward (#216). */
+  history_degraded_dense?: boolean;
   /** Codepoints not in the glyph atlas. A spike means users type glyphs we don't ship — widen ATLAS_PROFILE. */
   dropped_chars?: number;
   /** Top-20 dropped codepoints (U+HHHH keys) by frequency. Only present when dropped_chars > 0. */
@@ -305,6 +309,7 @@ export function toTrackEvent(ev: ProxyEvent): TrackEvent {
     if (info.historyFreezeStep !== undefined) out.history_freeze_step = info.historyFreezeStep;
     if (info.historyBudgetTrimmed) out.history_budget_trimmed = true;
     if (info.historyPackFill) out.history_pack_fill = true;
+    if (info.historyDegradedDense) out.history_degraded_dense = true;
     if (info.droppedChars !== undefined && info.droppedChars > 0) {
       out.dropped_chars = info.droppedChars;
     }
--- a/src/core/transform.ts
+++ b/src/core/transform.ts
@@ -51,6 +51,7 @@ import {
   ANTHROPIC_MAX_IMAGES,
   ANTHROPIC_HISTORY_IMAGE_BUDGET,
 } from './history.js';
+import type { HistoryCollapseInfo } from './history.js';
 import { noteHistoryRequest, recordFreezeStep } from './session-state.js';
 import type { GptHistoryOptions } from './openai-history.js';
 import { CACHE_CREATE_RATE, CACHE_READ_RATE } from './baseline.js';
@@ -812,6 +813,11 @@ export interface TransformInfo {
    *  didn't (exclude from rollup — cacheable=0 fallback is dishonest). 'failed': no
    *  baseline. undefined: no probe attempted. */
   baselineProbeStatus?: 'ok' | 'partial' | 'failed';
+  /** The model's legible history render (jetbrains-mono-14, #170) overflowed the
+   *  decoded image-byte budget, so the collapse was re-rendered at the dense
+   *  geometry and admitted instead of being abandoned to a raw-text forward
+   *  (#216). Diagnostic only. */
+  historyDegradedDense?: boolean;
 }
 
 // --- helpers ---------------------------------------------------------------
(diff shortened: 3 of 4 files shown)


ousamabenyounes:
Rebased onto current `main` (0.13.1) and verified with a real RED → GREEN locally (`vitest run tests/image-byte-budget.test.ts`, Node 24.15).

**RED** (fix in `src/core/transform.ts` + `src/core/tracker.ts` reverted, tests kept):
```
FAIL tests/image-byte-budget.test.ts > #216 > re-renders the collapse at dense instead of forwarding raw text (imaged-slab path)
FAIL tests/image-byte-budget.test.ts > #216 > re-renders the collapse at dense instead of forwarding raw text (text-slab path)
Tests  2 failed | 11 passed (13)
```
**GREEN** (with the fix):
```
Test Files  1 passed (1)
Tests  13 passed (13)
```

teamchong:
Thanks for this. I have concern on the approach: history is rendered legible on purpose. Tool results tolerate dense because they're mostly lookup, but history is where a misread turn compounds (lost instructions, wrong file, loops). Dense geometry hasn't been validated for that yet.

PR #235, opened 2026-08-14
fix(transform): classify Claude Code automode blocks as dynamic

eduardxdc:
## Problem

Claude Code automode injects `<cc_automode_session_rules>` and `<cc_automode_permissions>` blocks whose content evolves per turn (permissions/grants land mid-session). They were not in `DYNAMIC_BLOCK_TAGS`, so they sat in the static slab and re-keyed the imaged cache prefix on every turn — the worst cache shape (permanent create, never read). They also surfaced as a noisy `[pxpipe warn] unknown tag(s) in static slab: cc_automode_session_rules, cc_automode_permissions, severity, category` on every request.

## Fix

Add both automode blocks to `DYNAMIC_BLOCK_TAGS`, mirroring the existing `total_tokens` fix. Nested `<severity>`/`<category>` tags ride inside these blocks and stop being reported as unknown once their parents are classified dynamic.

## Tests

New `tests/dynamic-automode-tags.test.ts` (3 tests), mirroring `dynamic-total-tokens.test.ts`:
- static slab stays byte-identical when only the automode payload moves (`systemSha8` unchanged)
- automode tags are not reported in `unknownStaticTags`
- automode blocks route into the dynamic text (dynamic section differs across turns)

Verification: `pnpm typecheck` clean · 6/6 tests in the two dynamic-tag files pass · `pnpm build` succeeds with the tags in the bundle. (2 pre-existing failures in `image-byte-budget.test.ts` reproduce on clean upstream — canvas-render environment issue, unrelated.)

### DIFF
--- a/src/core/transform.ts
+++ b/src/core/transform.ts
@@ -902,6 +902,11 @@ const DYNAMIC_BLOCK_TAGS = [
   // paid for as a create and never read. It was showing up in
   // `unknownStaticTags`, which is exactly the canary that list exists to be.
   'total_tokens',
+  // Claude Code automode state — permissions/rules evolve per turn as grants
+  // land mid-session, so they must not bake into the cached image. Nested
+  // <severity>/<category> tags ride inside these blocks.
+  'cc_automode_session_rules',
+  'cc_automode_permissions',
 ] as const;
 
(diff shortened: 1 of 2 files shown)


teamchong:
Thanks! Superseded by #236

Issue #234, opened 2026-08-14
many 429 errors


Pyroarsonist:
I am using `pxpipe` with Claude Code and model `claude-sonnet-5`

Recently, out of nowhere, i started to have such errors:

```
[2026-08-14T08:47:46.916Z] POST /v1/messages → 429 (566ms tx=99ms up=467ms fb=480ms) compressed 112166ch → 5img/596559B
[pxpipe warn] unknown tag(s) in static slab: cc_automode_session_rules, cc_automode_permissions, severity, category  — may need to add to DYNAMIC_BLOCK_TAGS (per-turn) or KNOWN_STATIC_TAGS (static) in src/core/transform.ts
```

Requests are failing, nothing is proceeded.

Any suggestions what's going on?

When compressing is disabled, all requests going green 200.

teamchong:
can you share:

1. pxpipe --version and claude --version
2. From one failing request (DEBUG=true pxpipe warp -- claude): the first system block, and the 429 response body

suspect: your Claude Code build changed the identity block (the unknown cc_automode_* tags hint at a newer build), so pxpipe's exact-match lift misses it, it gets imaged, and requests lose subscription quota recognition → 429

Pyroarsonist:
```
$ npx pxpipe-proxy --version
0.13.1

$ claude --version
2.1.233 (Claude Code)
```

didn't found the system block or response body (i am using `npx pxpipe-proxy` instead)

i have those logs out of webpanel:

[logs.txt](https://github.com/user-attachments/files/31144449/scratch_7.txt)

hope that helps.

Pyroarsonist:
btw i am using ponytail and caveman too, maybe it would help too

teamchong:
made a private repo and added you: https://github.com/teamchong/pxpipe_issues_234

please run one failing session with

```sh
PXPIPE_DEBUG_CAPTURE_4XX=1 npx pxpipe-proxy warp -- claude
```

and upload ~/.pxpipe/4xx-bodies/*.json.gz and ~/.pxpipe/events.jsonl there (they can contain paths and request fragments, so not here). check the bodies for secrets first.

suspect the cc_automode_* blocks churn per turn and re-render the whole slab; #236 on main should fix that, next release.

PR #236, opened 2026-08-17
fix(transform): route cc_automode_*/severity/category into dynamic tail

parziva-1:
`cc_automode_session_rules`, `cc_automode_permissions`, `severity`, and `category` are per-turn
tag-shaped blocks that newer Claude Code builds inject into the system prompt (visible as
`unknownStaticTags` in `pxpipe stats`, which itself suggests adding entries here). None of the
four was in `DYNAMIC_BLOCK_TAGS`, so they fell into the static slab and got imaged along with the
rest of the identity block.

Related to #234: with `claude-sonnet-5` and compression on, requests started failing with 429 out
of nowhere, and the reported `stats`/`DEBUG` output showed these same unknown tags. This PR routes
the four tags into the dynamic tail — same mechanism `total_tokens` already uses, added above for
an analogous reason (see the comment on that entry).

I can't verify *why* the provider's subscription-quota recognition breaks when this block is
imaged — that's on the provider's side, not observable from here. This PR only fixes the
classification bug pxpipe's own telemetry already flagged, and reports what happened before/after
in my environment as real-world evidence, not just the unit test below:

- Model: `claude-sonnet-5`, Claude Code client `2.1.229`. `claude-fable-5,gemini-3.6-flash` are
  the built-in defaults; `claude-sonnet-5` added via `PXPIPE_MODELS` (opt-in, as documented).
- Render geometry: `CLAUDE_LEGIBLE_PROFILE` (172-col history strip, `jetbrains-mono-14`) — the
  default non-Fable Claude profile.
- Before this fix: 10/10 real requests through the proxy with `claude-sonnet-5` compression on
  failed with 429, all within a ~30s window — matches the shape reported in #234.
- After this fix: 9/9 real requests through the proxy (same model, same opt-in config) came back
  `200`, and `unknownStaticTags` no longer listed any of the four tags for those requests.
- Sample size is small (single machine, single session, one date: 2026-08-16/17). This is not a
  claim that it's the only cause of every 429 in #234 — only that it reproduces and resolves this
  specific mechanism. Leaving the broader call to the maintainer per CONTRIBUTING.md.

## Verify

```text
$ pnpm vitest run tests/dynamic-cc-automode-tags.test.ts
# before this diff (revert src/core/transform.ts to see it fail):
 ❯ tests/dynamic-cc-automode-tags.test.ts (3 tests | 3 failed)
     × does not report any of the four tags as unknown static blocks
     × keeps the static slab byte-identical when only the block content moves
     × routes the block into the dynamic text, not the imaged slab

# with this diff:
 ✓ tests/dynamic-cc-automode-tags.test.ts (3 tests | 3 passed)

$ pnpm test
 Test Files  1 failed | 72 passed (73)
      Tests  2 failed | 1154 passed (1156)
# The 2 failures are in tests/image-byte-budget.test.ts, unrelated to this change, and
# pre-exist on a clean `main` checkout — confirmed by stashing this diff and re-running the
# same file against unmodified main (same 2 failures, same assertions).

$ pnpm typecheck
# clean, no output
```

- [x] `pnpm test` and `pnpm typecheck` pass (the 2 pre-existing `image-byte-budget.test.ts`
      failures are unrelated to this change and reproduce on a clean `main` checkout — see Verify)

### DIFF
--- a/src/core/transform.ts
+++ b/src/core/transform.ts
@@ -902,6 +902,15 @@ const DYNAMIC_BLOCK_TAGS = [
   // paid for as a create and never read. It was showing up in
   // `unknownStaticTags`, which is exactly the canary that list exists to be.
   'total_tokens',
+  // Newer Claude Code identity/automode block (#234). Left in the static slab
+  // it gets imaged along with the rest of the prefix, and the provider's
+  // subscription-quota recognition on that block breaks -> spurious 429s on
+  // every compressed request. Route it to the dynamic tail instead, same as
+  // total_tokens above.
+  'cc_automode_session_rules',
+  'cc_automode_permissions',
+  'severity',
+  'category',
 ] as const;
 
 // Known-static slab tags — suppresses first-sighting `unknownStaticTags` noise
--- a/tests/dynamic-cc-automode-tags.test.ts
+++ b/tests/dynamic-cc-automode-tags.test.ts
@@ -1,0 +1,73 @@
+/**
+ * `<cc_automode_session_rules>`, `<cc_automode_permissions>`, `<severity>`, and
+ * `<category>` are per-turn blocks emitted by newer Claude Code builds. Left
+ * out of DYNAMIC_BLOCK_TAGS they fall into the static slab and get imaged
+ * along with the rest of the identity block (see #234).
+ *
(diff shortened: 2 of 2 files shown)


teamchong (approved):
Thanks

PR #281, opened 2026-09-16
fix(render): separate the ↵ marker with a blank cell on each side

### LINKED ISSUE #270: ↵ newline marker renders with no separating whitespace, so line-trailing values attach to the next line's identifier
The `↵` (U+21B5) hard-newline marker is rendered with **no separating whitespace on either side**, so the value ending one source line and the identifier starting the next fuse into a single visual token. A reader that transcribes the image perfectly still attributes line-trailing values to the *following* line.

This is a layout problem rather than a legibility one, which is why I'm filing it separately from the dense-render comprehension discussion in #160 — it is not fixed by a larger glyph cell or a better reader.

## What the render looks like

Source (240 lines, one per record):

```
L00002  ledger entry alpha bravo charlie delta echo foxtrot golf hotel india juliet 00014
L00003  ledger entry alpha bravo charlie delta echo foxtrot golf hotel india juliet 00021
```

Rendered strip, magnified 5x from the actual PNG:

```
…golf hotel india juliet 00014↵L00003  ledger entry alpha bravo…
```

Note the spacing hierarchy, which is inverted with respect to structure:

| boundary | separation in the render |
|---|---|
| between words inside a line | 1 space |
| between a record ID and its content | **2 spaces** |
| **between two source lines** (`↵`) | **0 spaces** |

The strongest boundary in the source gets the least visual separation of the three.

## Consequence

`00014↵L00003` reads as one compound token. In two independent reads at `temperature: 0` the model transcribed the marker as `/` and reported seeing pairs of the form `00042/L00043` and `00042/L00021`, then explained its answer in terms of that pairing — i.e. it treated the trailing number as belonging to the *following* identifier. Asked which five-digit number ends the line beginning `L00042` (ground truth `00294`), it answered `00047` and `00043`. Both are adjacent-line artefacts, not random noise.

The marker itself is present and the legend is emitted correctly in the banner ("The glyph ↵ (U+21B5) marks an original hard line break in content; treat it as a real newline"), so the mechanism is all there. It is defeated by the kerning.

## Why it may be worth more than its size suggests

For a coding agent this degrades gracefully — as noted in #160, the agent can fall back to `grep` when it cannot read something. But the failure here is *silent*: the read succeeds, it is simply off by one line. There is no signal that would prompt a fallback, and in a non-agentic pipeline (a plain API call with no filesystem) there is no fallback to reach for.

It only bites content where meaning lives in row alignment — tables, schedules, numbered clause lists, key/value dumps, log lines, code. Running prose is unaffected, since sentence boundaries survive reflow on punctuation alone.

## Reproduction

```python
lines = ["L%05d  ledger entry alpha bravo charlie delta echo foxtrot golf hotel india juliet %05d"
         % (i, i * 7 % 99991) for i in range(240)]
# send "\n".join(lines) as a system message through the proxy (~21.6k chars -> 1 image),
# then ask: "What five-digit number ends the line beginning L00042?"   (truth: 00294)
```

Observed on 0.13.2 via the OpenAI-compatible chat path; the rendered PNG is byte-identical on 0.13.0, so this is not a recent regression.

## Suggested fix

Pad the marker with a space on each side (` ↵ `) so it separates at least as strongly as an intra-line word gap. That costs one or two cells per source line against the wrap budget, which seems cheap next to a systematic off-by-one in line attribution. `markerScale` in `render.js` already reserves extra wrap budget for an enlarged marker, so the surrounding whitespace may be the smaller change of the two.

Happy to test a patch against the same fixture if that would help.

SulimanAbdulrazzaq:
The `↵` (U+21B5) hard-newline marker is laid out with no separating cell on
either side, so the value ending one source line and the identifier starting
the next fuse into a single visual token. The strongest boundary in the source
ends up with *less* separation than an intra-line word gap:

| boundary | before | after |
|---|---|---|
| between words inside a line | 1 cell | 1 cell |
| between a record ID and its content | 2 cells | 2 cells |
| between two source lines (`↵`) | **0 cells** | **1 cell each side** |

Before / after, same input (`L00002  ledger …00014` + `L00003  ledger …00021`):

```text
before:  …india juliet 00014↵L00003  ledger entry alpha…
after:   …india juliet 00014 ↵ L00003  ledger entry alpha…
```

## The fix

`padNewlineMarkers` is a pure string→string pass in the wrap layer, applied
next to `expandTabsInLine` and `escapeMissingGlyphs` in both `wrapLines` and
`measureContentCols`, so layout and canvas measurement stay in agreement and
the slot string keeps wrapping in lockstep with the text.

`reflow()` still emits the bare sentinel — the padding exists only on the
render copy, so the L0 `dereflow(reflow(text))` round-trip is untouched (the
first new test pins that).

## Cost

The padding spends 2 cells per source line of wrap budget. Measured rows for
the same input with and without the change:

| fixture | cols | before | after |
|---|---|---|---|
| 240 lines × ~100 chars | 312 | 70 | 71 |
| 240 lines × ~100 chars | 172 | 126 | 129 |
| 500 lines × ~8 chars | 312 | 14 | 17 |
| 500 lines × ~8 chars | 172 | 25 | 31 |

So ~1–2% on log/code/prose-shaped lines, and up to ~24% on pathologically
short ones. `NL_SENTINEL_PAD_CELLS` is a single constant if you want to tune
or gate that.

I have deliberately not made any claim about what this does to model
readability — that needs per-model runs I can't hold to the standard in
CONTRIBUTING. This PR only fixes the layout invariant, which a test can check.

## Verify

```text
$ npx vitest run tests/reflow.test.ts tests/paging.test.ts tests/glyph-escape.test.ts \
    tests/history.test.ts tests/openai-history.test.ts tests/docs-integrity.test.ts
 Test Files  6 passed (6)
      Tests  186 passed (186)

$ pnpm typecheck
(clean, exit 0)
```

Reverting just the `padNewlineMarkers` body to `return line` fails 3 of the 5
new tests, so they do pin the behaviour.

Full `pnpm test` on my machine has a few timeout-only flakes unrelated to this
change (`node-security`, `export-git-e2e`, `proxy-usage`, and the
`~/.claude/projects` corpus scan in `reflow.test.ts`); all of them pass when
run on their own, both with and without this patch.

Fixes #270

- [x] `pnpm typecheck` passes; `pnpm test` passes except the pre-existing timeout-only flakes noted above (all green when run individually, with and without this patch)


SulimanAbdulrazzaq:
@teamchong when you have a moment, could you approve the workflow run and take a look? The `↵` hard-newline marker is laid out with no blank cell on either side, so the value that ends one source line and the identifier that starts the next fuse into one token; this puts one blank cell on each side. CI is waiting for approval, and the branch has no conflicts.

